PR #10, opened 2020-10-14
Extend sysbox-fs to support overlayfs mounts on unprivileged user-ns

rodnymolina:
This one is a key requirement to support Sysbox on Linux Distributions other than Ubuntu.

Our goal is to allow overlay file-systems to be mounted from within sys containers (i.e. unprivileged user-ns contexts) without the need to rely on external software components such as fuse-overlayfs. Refer to issue #83 for more details: nestybox/sysbox#83.

### DIFF
--- a/nsenter/event.go
+++ b/nsenter/event.go
@@ -620,6 +620,10 @@ func (e *NSenterEvent) processMountSyscallRequest() error {
 
 	payload := e.ReqMsg.Payload.([]domain.MountSyscallPayload)
 
+	// For overlayfs mounts we adjust 'nsexec' process' personality (i.e.
+	// uid/gid and capabilities) to match the one of the original process
+	// performing the syscall. Our goal is mainly to avoid permission issues
+	// while accessing kernel's created overlayfs components.
 	if payload[0].FsType == "overlay" {
 		// Create a 'process' struct to represent the 'sysbox-fs nsenter' process
 		// executing this logic.
--- a/nsenter/event.go
+++ b/nsenter/event.go
@@ -39,6 +39,7 @@ import (
 
 	"github.com/nestybox/sysbox-fs/domain"
 	"github.com/nestybox/sysbox-fs/fuse"
+	"github.com/nestybox/sysbox-fs/process"
 	"github.com/nestybox/sysbox-runc/libcontainer"
 )
 
@@ -83,6 +84,9 @@ type NSenterEvent struct {
 
 	// Zombie Reaper (for left-over nsenter child processes)
 	reaper *zombieReaper
+
+	// Backpointer to Nsenter service
+	service *nsenterService
 }
 
 //
@@ -616,6 +620,33 @@ func (e *NSenterEvent) processMountSyscallRequest() error {
 
 	payload := e.ReqMsg.Payload.([]domain.MountSyscallPayload)
 
+	if payload[0].FsType == "overlay" {
+		// Create a 'process' struct to represent the 'sysbox-fs nsenter' process
+		// executing this logic.
+		process := e.service.prs.ProcessCreate(0, 0, 0)
+
+		// Extract payload-header from the first element of the payload slice.
+		header := payload[0].Header
+
+		// Adjust 'nsenter' process personality to match the end-user's original
+		// process.
+		if err := process.Camouflage(
+			header.Uid,
+			header.Gid,
+			header.Root,
+			header.Cwd,
+			header.Capabilities); err != nil {
+
+			// Send an error-message response.
+			e.ResMsg = &domain.NSenterMessage{
+				Type:    domain.ErrorResponse,
+				Payload: &fuse.IOerror{RcvError: err},
+			}
+
+			return nil
+		}
+	}
+
 	// Perform mount instructions.
 	for i = 0; i < len(payload); i++ {
 		err = unix.Mount(
@@ -626,18 +657,12 @@ func (e *NSenterEvent) processMountSyscallRequest() error {
 			payload[i].Data,
 		)
 		if err != nil {
-			// Create error response msg.
-			e.ResMsg = &domain.NSenterMessage{
-				Type:    domain.ErrorResponse,
-				Payload: &fuse.IOerror{RcvError: err},
-			}
-
 			break
 		}
 	}
 
 	if err != nil {
-		// Unmount previously executed mount instructions (unless it's a remount)
+		// Unmount previously executed mount instructions (unless it's a remount).
 		//
 		// TODO: ideally we would revert remounts too, but to do this we need information
 		// that we don't have at this stage.
@@ -645,6 +670,12 @@ func (e *NSenterEvent) processMountSyscallRequest() error {
 			if payload[j].Flags&unix.MS_REMOUNT != unix.MS_REMOUNT {
 				_ = unix.Unmount(payload[j].Target, 0)
 			}
+		}
+
+		// Create error response msg.
+		e.ResMsg = &domain.NSenterMessage{
+			Type:    domain.ErrorResponse,
+			Payload: &fuse.IOerror{RcvError: err},
 		}
 
 		return nil
@@ -890,7 +921,11 @@ func Init() (err error) {
 	// specific env vars.
 	os.Clearenv()
 
-	var event = NSenterEvent{}
+	var nsenterService nsenterService
+	var processService = process.NewProcessService()
+
+	nsenterService.Setup(processService)
+	var event = NSenterEvent{service: &nsenterService}
 
 	// Process incoming request.
 	err = event.processRequest(pipe)
--- a/nsenter/event.go
+++ b/nsenter/event.go
@@ -625,7 +625,8 @@ func (e *NSenterEvent) processMountSyscallRequest() error {
 		// executing this logic.
 		process := e.service.prs.ProcessCreate(0, 0, 0)
 
-		// Extract payload-header from the first element of the payload slice.
+		// Extract payload-header from the first element (just one in overlayfs
+		// mounts) of the payload slice .
 		header := payload[0].Header
 
 		// Adjust 'nsenter' process personality to match the end-user's original
--- a/process/process.go
+++ b/process/process.go
@@ -61,52 +61,72 @@ func (ps *processService) ProcessCreate(
 }
 
 type process struct {
-	pid      uint32                  // process id
-	root     string                  // root dir
-	cwd      string                  // current working dir
-	uid      uint32                  // effective uid
-	gid      uint32                  // effective gid
-	sgid     []int                   // supplementary groups
-	cap      cap.Capabilities        // process capabilities
-	status   map[string]string       // process status fields
-	nsInodes map[string]domain.Inode // process namespace inodes
-	ps       *processService         // pointer to parent processService
+	pid         uint32                  // process id
+	root        string                  // root dir
+	hroot       string                  // root dir from host perspective
+	cwd         string                  // current working dir
+	hcwd        string                  // current working dir from host perspective
+	uid         uint32                  // effective uid
+	gid         uint32                  // effective gid
+	sgid        []int                   // supplementary groups
+	cap         cap.Capabilities        // process capabilities
+	status      map[string]string       // process status fields
+	nsInodes    map[string]domain.Inode // process namespace inodes
+	initialized bool                    // process initialization completed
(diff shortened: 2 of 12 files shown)


nsenter/event.go
```diff
@@ -616,6 +620,34 @@ func (e *NSenterEvent) processMountSyscallRequest() error {
 
 	payload := e.ReqMsg.Payload.([]domain.MountSyscallPayload)
 
+	if payload[0].FsType == "overlay" {
```
ctalledo, line 623: Let's add a comment indicating why specifically for overlayfs mounts the handler needs to adopt the uid:gid of the container process that performed the mount syscall.

process/process.go
```diff
@@ -61,34 +61,102 @@ func (ps *processService) ProcessCreate(
...
+	status      map[string]string       // process status fields
+	nsInodes    map[string]domain.Inode // process namespace inodes
+	initialized bool                    // process initialization completed
+	ps          *processService         // pointer to parent processService
 }
 
 func (p *process) Pid() uint32 {
+
+	if !p.initialized {
+		p.getInfo()
```
ctalledo, line 82: It looks like p.getInf() should be renamed p.init() ? Because it appears it's initializing the process struct.

process/process.go
```diff
@@ -61,34 +61,102 @@ func (ps *processService) ProcessCreate(
...
+func (p *process) Root() string {
+
+	if !p.initialized {
+		p.getInfo()
+	}
+
+	return p.root
+}
+
+func (p *process) IsSysAdminCapabilitySet() bool {
```
ctalledo, line 124: The IsXCapabilitySet() wrappers don't add much value in my view. The code is equally readable using p.isCapabilitySet(cap.EFFECTIVE, cap.X). Plus there are many caps, and we are not going to have wrappers for all.

process/process.go
```diff
@@ -133,6 +201,76 @@ func (p *process) initCapability() error {
 	return nil
 }
 
+// Camouflage() method's purpose is to adjust the process personality to another
+// process. The ultimate goal is to be able to utilize a 'proxy' process (such
+// as 'sysbox-fs nsenter') to execute instructions on behalf of the original
+// process.
+func (p *process) Camouflage(
```
ctalledo, line 208: I would call it "SetPersonality"; "camouflage" sounds like it's hiding something. Just a suggestion.

process/process.go
```diff
@@ -133,6 +201,76 @@ func (p *process) initCapability() error {
...
+	}
+
+	// Initialize capability struct if not already done.
+	if p.cap == nil {
+		if err := p.initCapability(); err != nil {
+			return err
+		}
+	}
+
+	// If UID corresponds to 'root' user then adjust capabilities of this
```
ctalledo, line 230: Do you mean "non-root" in the container or at host level?

Also, why do we differentiate between root and non-root? Why isn't the setting of the personality identical for both cases?

process/process.go
```diff
@@ -280,6 +418,10 @@ func (p *process) PathAccess(path string, aMode domain.AccessMode) error {
 // getInfo retrieves info about the process
 func (p *process) getInfo() error {
 
+	if p.initialized {
```
ctalledo, line 421: I think p.getInfo() is really p.init(). 

p.getInfo() sounds like "return me the process info", but in fact in does not. It initialized the process object.

Also: do we need locking in the process object?

process/process.go
```diff
@@ -403,9 +550,9 @@ func (p *process) pathAccess(path string, mode domain.AccessMode) error {
 	// Determine the start point.
 	var start string
 	if filepath.IsAbs(path) {
-		start = p.root
+		start = p.hroot
```
ctalledo, line 553: Question: why p.hroot and not p.root?

seccomp/mount.go
```diff
@@ -132,6 +135,11 @@ func (m *mountSyscallInfo) process() (*sysResponse, error) {
 func (m *mountSyscallInfo) processProcMount(
 	mip *mountInfoParser) (*sysResponse, error) {
 
+	// Adjust mount attributes attending to process' root path.
+	if err := m.rootAdjust(); err != nil {
```
ctalledo, line 139: What is rootAdjust() doing? Making the path's absolute? or more ...

seccomp/mount.go
```diff
@@ -181,8 +189,10 @@ func (m *mountSyscallInfo) createProcPayload(
 	// apply this to the new mountpoint (otherwise we will get a permission
 	// denied from the kernel when doing the mount).
 	procInfo := mip.GetInfo("/proc")
-	if _, ok := procInfo.VfsOptions["ro"]; ok == true {
-		payload[0].Flags |= unix.MS_RDONLY
+	if procInfo != nil {
```
ctalledo, line 192: When is procInfo null? And what do we do when it is null?

seccomp/mount.go
```diff
@@ -528,6 +620,13 @@ func (m *mountSyscallInfo) createRemountPayload(
 func (m *mountSyscallInfo) processBindMount(
 	mip *mountInfoParser) (*sysResponse, error) {
 
+	// Notice that we are not adjusting mount attributes to deal with chroot'ed
```
ctalledo, line 623: I don't fully understand this. Let's talk.

seccomp/mount.go
```diff
@@ -604,6 +703,24 @@ func (m *mountSyscallInfo) createBindMountPayload(
...
+// so that sysbox-fs can carry out the mount in the expected context.
+func (m *mountSyscallInfo) rootAdjust() error {
+
+	root := m.syscallCtx.root
+
+	if root == "/" {
+		return nil
+	}
+
+	// 'Target' attribute will always require adjustment in chroot'ed envs.
```
ctalledo, line 722: How about the source attribute?

seccomp/umount.go
```diff
@@ -63,7 +64,12 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
 		// This restriction should be fine because unmounting /proc inside the
 		// container is not a useful thing to do anyways. Same rationale applies
 		// to "/sys".
-		if u.Target == "/proc" || u.Target == "/sys" {
+		//
+		// Also, notice that we want to clearly differentiate the root /proc and
+		// /sys mounts from those that are present within 'chroot'ed contexts. In
+		// the later case we want to allow users to mount (and umount) both /proc
+		// and /sys file-systems.
+		if (u.Target == "/proc" || u.Target == "/sys") && (u.syscallCtx.root == "/") {
```
ctalledo, line 72: Should the u.Target == "/sys" check be moved under case "sysfs" below?

rodnymolina: Why?

The idea is to decide, as early as possible, if sysbox-fs is handling the umount or not -- this is critical-path during container initialization & removal.

seccomp/mount.go
rodnymolina: Excluding bind-mount operations, can you think about a scenario that would require "source" attribute to hold a FS path in the operations carried over by sysbox-fs? I'm asking coz prepending a path is not trivial in the "source" case, as this field is somewhat overloaded (FS types, FS paths, etc).

See my previous paragraph about how kernel is obviating "root" attribute in chroot'ed contexts -- that's why i'm obviating the bind-mount case.

seccomp/mount.go
rodnymolina: Bind-mounts are (apparently) not properly managed by kernel in chroot'ed contexts. See below that bind-mounts done within the chroot jail are only displayed outside the chroot ...

My point here is that we should keep consistency with what the kernel is currently doing, so that's why i'm not calling rootAdjust() method.

```
rmolina@dev-vm1:~$ docker run --privileged -it --rm --name ubuntu ubuntu
root@afa472988d2d:/#

root@afa472988d2d:/# pwd
/
root@afa472988d2d:/# mkdir rootfs
root@afa472988d2d:/# cp -a /usr rootfs/
root@afa472988d2d:/# cp -a /lib rootfs/
root@afa472988d2d:/# cp -a /lib64 rootfs/
root@afa472988d2d:/# cp -a /bin rootfs/

root@afa472988d2d:/# chroot /rootfs
#
# mount
mount: /proc: mount point does not exist.

# mkdir /proc
# mount -t proc proc /proc

# findmnt
TARGET SOURCE FSTYPE OPTIONS
/proc  proc   proc   rw,relatime
#

# mkdir -p /a /a/b /x /x/y
# mount -o bind /a /x
#
# findmnt
TARGET SOURCE FSTYPE OPTIONS
/proc  proc   proc   rw,relatime
#
```

<-- Outside chroot jail (back to container):

```
root@afa472988d2d:/# findmnt
TARGET                                SOURCE                                                                           FSTYPE  OPTIONS
/                                     overlay                                                                          overlay rw,relatime,lowerdir=/var/lib/docker/overlay2/l/6ZJX5TKOFYRQRL2O3E7R2I65S
|-/rootfs/proc                        proc                                                                             proc    rw,relatime
|-/rootfs/x                           overlay[/rootfs/a]                                                               overlay rw,relatime,lowerdir=/var/lib/docker/overlay2/l/6ZJX5TKOFYRQRL2O3E7R2I65S
```

seccomp/mount.go
rodnymolina: No, it's not converting the path to absolute (it is already absolute by now -- prior routine takes care of it), rootAdjust() is simply adjusting the "target" mount parameter to account for an offset in the process' root path attribute.

seccomp/mount.go
rodnymolina: This is the fix for issue [#82](https://github.com/nestybox/sysbox/issues/82). Basically, we were assuming that a procfs is always present, which is not necessarily the case in chroot jails.

If there's no procfs (procinfo == null) we just keep on going w/o dereferencing the null pointer.

process/process.go
rodnymolina: See that i'm not changing the behavior here. Before this PR, "p.root" held the path as seen from the host (/proc/<pid>/root). Now that same value is carried by the "p.hroot" parameter. I'm now utilizing "p.root" for the process' root attribute.

process/process.go
rodnymolina: I like p.init() idea, will change.

No, no locking needed as operations over any given process object happen within a unique execution context.

process/process.go
rodnymolina: Not sure about this one, i think i like "camouflage" more than "setPersonality". Will think about it.

process/process.go
rodnymolina: Reason for these wrappers is to avoid introducing an external dependency ("cap" type, which seats in sysbox-libs/capabilities/capability.go) into sysbox-fs' domain layer. This layer should be kept clean of external components (other than go stdlib pkgs).

process/process.go
rodnymolina: Yep, agree, will change.

process/process.go
rodnymolina: The idea here is to make this method generic enough (beyond our current 'nsenter' use-case) so that it can serve any incoming UID. In the scenario where UID==0 there's no need to execute setresuid/gid.

Putting the use-case aside, i now realize that this "if" statement is missing a "&& gid == 0" condition for it to be correct.

nsenter/event.go
rodnymolina: done.

seccomp/umount.go
rodnymolina: "u.syscallCtx.root" variable is correctly pointing to "/" to refer to paths to be mount (or mounted) in the host system. See that "root" attribute holds the process' root-path from a host perspective, not from a process one.

seccomp/mount.go
rodnymolina: I ran this experiment again. Kernel is indeed making the bind-mount in the expected jailed context, but for some reason 'findmnt' is not displaying them. However, /proc/pid/mountinfo has the right information, so looks like a 'findmnt' bug.

My point: we need to handle bind-mounts within chroot jails.

seccomp/mount.go
rodnymolina: As mentioned above, sysbox-fs needs to deal with bind-mounts inside chroot jails, so yes, we need to take "source" attribute in consideration too.

process/process.go
rodnymolina: Changed to "adjustPersonality".

ctalledo (approved):

PR #13, opened 2020-11-25
Fix for sysbox issue #130 (sysbox-fs portion)

ctalledo:


domain/nsenter.go
```diff
@@ -179,3 +181,9 @@ type UmountSyscallPayload struct {
 	FsType uint8  `json:"-"`
 	Flags  uint64 `json:"flags"`
 }
+
+type ChownSyscallPayload struct {
+	Target string `json:"target"`
+	Uid    int    `json:"uid"`
```
rodnymolina, line 187: I find these Uid/Gid fields somewhat confusing as we already have equivalent fields for the process. I'd rather have these call differently, i.e. cUid, cGid.

ctalledo: I think the structure sets the context for the fields: in this case it's a chown syscall payload struct, so it's clear the uid/gid fields refer to the chown uid/gid, not the process uid/gid. I could call them targetUid and targetGid if that helps.

rodnymolina: I'm referring to event.go file. There we are using Uid/Gid fields in a couple of sections while having Uid/Gid carrying different semantics. Also, we may need to include the 'header' struct (where there's already a Uid/Gid) inside this message in the future.

seccomp/chown.go
```diff
@@ -0,0 +1,205 @@
+//
+// Copyright 2019-2020 Nestybox, Inc.
```
rodnymolina, line 2: minor: new files should just expose current year.

seccomp/chown.go
```diff
@@ -0,0 +1,205 @@
...
+// sys container would fail without this code is that /sys is owned by the
+// host's true root, so it shows up as "nobody:nogroup" inside the sys container
+// and thus its ownership can't be changed from within the container. Some apps
+// running inside the container (e.g,. RedHat's RPM package manager) want to
+// chown /sys to root:root, causing the apps to get an EPERM and fail.  As a
+// work-around, Sysbox ignores chown to "/sys" inside the sys container (or in
+// any inner containers). All other chown operations are handled normally by the
+// kernel.
+
+func (t *syscallTracer) ignoreChown(cntr domain.ContainerIface, req *sysRequest, absPath string) bool {
```
rodnymolina, line 40: This comment applies to the entire file.

syscallTracer type and its methods is fully defined in tracer.go. Please let's keep it like that for proper organization/encapsulation. See how we've done it for all other syscalls (i.e. you can use object 'chwownSyscallInfo' instead).

rodnymolina (changes requested):
Looks good. A couple of changes required.

seccomp/chown.go
```diff
@@ -0,0 +1,205 @@
...
+
+	mi := mip.GetInfo(absPath)
+	if mi == nil || mi.FsType != "sysfs" {
+		return false
+	}
+
+	return true
+}
+
+func (t *syscallTracer) processChown(
```
rodnymolina, line 70: Looks like this entire function can be moved to tracer.go as its coupled to syscallTracer types such as sysRequest. Ideally, chown implementation shouldn't need to know anything about our particular syscallTracer implementation.

seccomp/chown.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	if t.ignoreChown(cntr, req, path) {
+		logrus.Debugf("Ignoring fchown syscall from pid %d: path = %v, uid = %v, gid = %v", req.Pid, path, uid, gid)
+		return t.createSuccessResponse(req.Id), nil
+	}
+
+	return t.createContinueResponse(req.Id), nil
+}
+
```
rodnymolina, line 132: Same as above.

domain/nsenter.go
ctalledo: I changed it to TargetUid and TargetGid.

seccomp/chown.go
ctalledo: Done.

seccomp/chown.go
ctalledo: Done.

seccomp/chown.go
ctalledo: Done.

seccomp/chown.go
ctalledo: Done.

seccomp/mount.go
```diff
@@ -169,6 +169,51 @@ func (m *mountSyscallInfo) processProcMount(
...
+	// chown, then remount read-only. This would in turn require 3 nsenter
+	// events, because the namespaces that we must enter for each are not the
+	// same (in particular for the chown to succeed, we must not enter the
+	// user-ns of the container).
+
+	if m.Flags&unix.MS_RDONLY == unix.MS_RDONLY {
+		return m.tracer.createSuccessResponse(m.reqId), nil
+	}
+
+	chownPayload := createChownPayload(m.Target, m.uid, m.gid)
```
rodnymolina, line 188: The rest of the code in this function is specific to Chown handling operations, meaning, you could place it in a ChownSyscallInfo.Process() method. That way you avoid leaving chown-specific logic in this file, which would need to be replicated in chownSyscallInfo class later on.

```
	chown := &chownSyscallInfo{
		syscallCtx: syscallCtx{
			syscallNum: int32(req.Data.Syscall),
			reqId:      req.Id,
			pid:        req.Pid,
			cntr:       cntr,
			tracer:     t,
		},
		path:     path,
		ownerUid: uid,
		ownerGid: gid,
	}

return chown.Process()
```

rodnymolina (approved):
Looks good! I have one last suggestion though.

PR #19, opened 2020-12-10
Refactor of sysbox-fs common handlers.

ctalledo:
This commit makes changes that clarify the purpose of the common handlers in
sysbox-fs. Specifically, it makes the following changes:

* Renames common handler -> procSysCommon handler.

* Renames the sysCommon handler to sys handler.

* Cleans up comments in the /proc and /sys handlers.

* Causes the proc handler to handle all accesses under
  /var/lib/sysboxfs/<container-id>/proc/ for which no other handler exists (just
  like the sys handler does).


rodnymolina (approved):
Nothing to add. LG!

PR #20, opened 2020-12-11
Improve sysbox-fs handlers to deal with concurrent accesses to emulated resources.

ctalledo:
It's possible that within a sys container, multiple processes may be accessing
the same sysbox-fs backed resources under procfs and sysfs (e.g., /proc/swaps,
/proc/sys/net/netfilter/nf_conntrack_max, etc). To ensure consistency across
these accesses, sysbox-fs needs to implement per-resource, per-container
locking. Such locking partially existed, but was not sufficient. This commit
fixes this.

In addition, some of the sysbox-fs handlers that emulate procfs and sysfs
resources must sometimes write to the kernel's real procfs and sysfs.  In this
case, an additional per-resource locking across all containers is required, to
ensure sysbox-fs writes consistent values to the real procfs or sysfs.


handler/implementations/maxIntBaseHandler.go
```diff
@@ -212,6 +218,11 @@ func (h *MaxIntBaseHandler) fetchFile(
 	n domain.IOnodeIface,
 	c domain.ContainerIface) (string, error) {
 
+	// We need the per-resource lock since we are about to access the resource on
+	// the host FS. See pushFile() for a full explanation.
+	h.Lock.Lock()
+	defer h.Lock.Unlock()
```
rodnymolina, line 224: Let's return after the if-errors instead to reduce critical section, logrus is known to be slow.

handler/implementations/maxIntBaseHandler.go
```diff
@@ -256,6 +271,9 @@ func (h *MaxIntBaseHandler) pushFile(
 	// agents that write to the same sysctl. That's because there is no guarantee
 	// that the other host agent will read-after-write and retry as sysbox does.
 
+	h.Lock.Lock()
```
rodnymolina, line 274: I'm a bit concerned that we will eventually introduce a bug in this code that will end up in deadlock. Can you add a note somewhere (not sure where's the proper location: handler struct definition or container one), to define the lock-acquisition protocol that we'll follow? container-ext-lock -> container-int-lock -> handler-lock.

handler/implementations/procSysCommon.go
```diff
@@ -248,10 +250,14 @@ func (h *ProcSysCommonHandler) Write(
 	// If caching is enabled, store the data in the cache and do a write-through to the
 	// host FS. Otherwise just do the write-through.
 	if h.Cacheable && domain.ProcessNsMatch(process, cntr.InitProc()) {
+
+		cntr.Lock()
```
rodnymolina, line 254: Mmm, for this particular handler this lock could end up acting as a giant-lock if the sys-container is (e.g.) a k8s node, specially if we take into account nsenter's latency. We may need to think about an alternate approach for this in the future. Perhaps we need to partition the /proc/sys hierarchy in various common handlers.

rodnymolina (changes requested):
Minor request change and a couple of items to talk about.

state/container.go
```diff
@@ -42,64 +41,66 @@ type container struct {
 	dataStore     domain.StateDataMap               // Handler's container-specific storage blob
 	initProc      domain.ProcessIface               // container's init process
 	service       domain.ContainerStateServiceIface // backpointer to service
+	intLock       sync.RWMutex                      // internal lock
```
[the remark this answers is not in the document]
ctalledo: The answer is "no", meaning that we could do with the external lock only. But I think removing the internal lock is error prone because it means that all users of the container type must now explicitly lock. This makes it harder to maintain in my view.

handler/implementations/maxIntBaseHandler.go
ctalledo: Sure, done.

handler/implementations/maxIntBaseHandler.go
ctalledo: Good idea, I added the text in the description of the handlerBase object.

// HandlerBase is a type common to all handlers                                                                                                                                                                                                                                                                               
//                                                                                                                                                                                                                                                                                                                            
// Note: the "Lock" variable can be used to synchronize across concurrent                                                                                                                                                                                                                                                     
// executions of the same handler (e.g., if multiple processes within the same                                                                                                                                                                                                                                                
// sys container or across different sys containers are accessing the same                                                                                                                                                                                                                                                    
// sysbox-fs emulated resource). When obtaining the handler lock, only do so to                                                                                                                                                                                                                                               
// synchronize accesses to host resources associated with the emulated resources                                                                                                                                                                                                                                              
// (e.g., if a handler needs to write to the host's procfs for example). While                                                                                                                                                                                                                                                
// holding the handler lock, avoid accessing objects of "container" struct type as                                                                                                                                                                                                                                            
// those have a dedicated lock which is typically held across invocations of the                                                                                                                                                                                                                                              
// handler lock. Violating this rule may result in deadlocks.

handler/implementations/procSysCommon.go
ctalledo: Note that it's a per-container lock, so it does not slow down accesses to `proc/sys` from processes in different containers. Let's see if this causes problems from a performance perspective (e.g., if it slows down the k8s tests) and go from there.

state/container.go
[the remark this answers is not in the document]
rodnymolina: Yes, i agree that 1. may be less error-prone with the current complexity (low). But if we see an upward trend in concurrency-handling logic, we should consider switching to message-passing approach so that we can sleep at night. Let's go with this for now.

handler/implementations/procSysCommon.go
rodnymolina: I understand, that's why i mentioned a 'large' system-container.

rodnymolina (approved):
LG

PR #21, opened 2020-12-23
Add sysbox-fs command line option to write logs using json format.

ctalledo:
This commit adds a command line option to sysbox-fs
to log using JSON. This is helpful with tools that parse
logging.

In addition, this change modified the default text logging
to be done without colors, unless it's done to a TTY
(in which case coloring is enabled).

### DIFF
--- a/cmd/sysbox-fs/main.go
+++ b/cmd/sysbox-fs/main.go
@@ -179,6 +179,11 @@ func main() {
 			Value: "info",
 			Usage: "log categories to include (debug, info, warning, error, fatal)",
 		},
+		cli.StringFlag{
+			Name:  "log-format",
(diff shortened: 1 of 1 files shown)


rodnymolina (commented):
Where's the logic to do this?

"unless it's done to a TTY (in which case coloring is enabled)."

ctalledo:
> Where's the logic to do this?

This is part of the default behavior of the Logrus package that Sysbox uses for logging. See here:

https://github.com/sirupsen/logrus

"Nicely color-coded in development (when a TTY is attached, otherwise just plain text)"

rodnymolina:
Got it.

rodnymolina (approved):
LG

PR #24, opened 2021-01-11
Mount hardening changes

rodnymolina:
There are multiple commits as part of this PR. Our main goal here is to enhance the overall system security by preventing actions executed within system containers to have any impact on the host file-system. As part of these changes, we have also refactored sysbox-fs mountInfo parser and associated logic.


seccomp/umount.go
```diff
@@ -117,15 +124,81 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	// There must be mountinfo state present for this target. Otherwise, let
+	// kernel handle the error.
+	info := mip.GetInfo(u.Target)
+	if info == nil {
+		return true, u.tracer.createContinueResponse(u.reqId)
+	}
+
+	// Return 'eperm' if the umount operation is over an immutable mountpoint.
+	if u.cntr.IsImmutableMountID(info.MountID) {
+		return false, u.tracer.createErrorResponse(u.reqId, syscall.EPERM)
```
ctalledo, line 144: I am wondering if we should return EINVAL (target is not a mountpoint) instead of EPERM. The reason is that in the future we may hide immutable mounts inside the container, such that a process inside the container won't be able to know that these are mounts but will instead see them as regular files/directories. If that's the case, the EINVAL would be a more appropriate error.

ctalledo: Also: this means we return an error for unmounts of /proc as well as the subdirs of /proc backed by sysbox-fs (e.g., /proc/sys). It's OK to return the error on the unmount of /proc itself, but for the subdirs we should ignore the unmount, as described in line 37 of this same file.

ctalledo: Do we consider `/proc` and it's sysbox-fs backed sub-dirs (e.g., `/proc/sys`) as immutable here? if so, I have two comments:

1) We want to ignore unmounts subdirs of `/proc` backed by sysbox-fs. See line 37 of this same file for the rationale. This change would break this.

2) Why would we need the `if-else` in lines 65 and 117 of this file then? They would not be reachable it seems.

state/containerDB.go
```diff
@@ -151,6 +157,14 @@ func (css *containerStateService) ContainerRegister(c domain.ContainerIface) err
 		)
 	}
 
+	// A per-container mountInfoParser object will be created here to hold the
+	// mount-state created by sysbox-runc during container initialization.
+	mip, err := css.mts.NewMountInfoParser(currCntr, c.InitPid(), true)
```
ctalledo, line 162: Question: do we need "deep" parsing here? Or could we do with the shallower parsing (to make the container registration faster)?

mount/infoParser.go
```diff
@@ -472,3 +450,109 @@ func (mi *mountInfoParser) HasNonSysboxfsSubmount(basemount string) bool {
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// IsBindMount verifies if the passed mountinfo entry is a bind-mount.
+func (mi *mountInfoParser) IsBindMount(info *domain.MountInfo) bool {
+
```
ctalledo, line 496: Suggestion: let's add a comment here with a high level description on how we discern that the given mount is a bind-mount. That way we won't have to re-think this in the future.

mount/infoParser.go
```diff
@@ -159,13 +127,23 @@ func (mi *mountInfoParser) parseData(data []byte) error {
 
 		mi.mpInfo[parsedMounts.MountPoint] = parsedMounts
 		mi.idInfo[parsedMounts.MountID] = parsedMounts
+
+		// Optimization needed here -- this logic makes sense only for remount ops.
```
ctalledo, line 131: This comment is a bit confusing; at this level nothing is known about remount ops.

mount/infoParser.go
```diff
@@ -53,78 +53,46 @@ import (
...
-	cntr   domain.ContainerIface
-	pid    uint32
-	deep   bool                  // superficial vs deep parsing mode
-	mpInfo map[string]*mountInfo // mountinfo, indexed by path
-	idInfo map[int]*mountInfo    // mountinfo, indexed by mount ID
+	mh        *mountHelper
+	cntr      domain.ContainerIface
+	pid       uint32
+	deep      bool                           // superficial vs deep parsing mode
+	mpInfo    map[string]*domain.MountInfo   // mountinfo, indexed by path
```
ctalledo, line 67: "indexed by path" -> "indexed by mountpoint path" (since mounts also have a root path).

seccomp/umount.go
ctalledo: Also: how come we are not checking for `(u.syscallCtx.root == "/")` here (see L79 for an example of when we use this check)? Don't we need to know that the process calling this has a root matching that of the sys container's init process?

mount/infoParser.go
rodnymolina: Agree.

mount/infoParser.go
rodnymolina: Yep, that's an old comment I forgot to remove.

seccomp/umount.go
```diff
@@ -117,15 +124,81 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	}
+
+	// At this point we have already concluded that the mountpoint being
+	// unmounted doesn't match any of the immutable resources, so we could
```
ctalledo, line 148: "doesn't match any of the immutable resources, ..." -> "doesn't match any of the immutable resources by mount ID, ...

seccomp/umount.go
```diff
@@ -117,15 +124,81 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	// be tempted to allow all umount instructions reaching this stage to
+	// succeed. However, there is one particular case that we must identify
+	// to prevent the exposure of immutable masked resources. This is the case
```
ctalledo, line 151: "This is the case of an umount instruction launched from a different mount namespace targeting an immutable resource (e.g., unshare -m umount /var/lib/docker), because that mount namespace has a different set of mount IDs. The rest of the code ..."

seccomp/umount.go
```diff
@@ -117,15 +124,81 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	// of an umount instruction launched from an unshare() context that targets
+	// an immutable resource. The rest of the code in this function deals with
+	// the identification and handling of this particular case.
+
+	p := u.processData
+
+	// Allow umount if this one is launched from a process whose root-inode
```
ctalledo, line 158: We have a bug here: just because the root is different, it does not mean it's not referencing the same mount. For example, say a process inside the sys container is in a dedicated mount namespace has a root of `/var/lib/`; if that process decides to unmount `/docker`, it's really unmounting `/var/lib/docker` inside the sys container, so it should not be allowed (since `/var/lib/docker` is an immutable mount). But this code will allow it.

seccomp/umount.go
```diff
@@ -117,15 +124,81 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	// differs from the sys container's initPid one.
+	if p.RootInode() != u.cntr.InitProc().RootInode() {
```
ctalledo, line 160: We have a bug here: just because the root is different, it does not mean it's not referencing the same mount. For example, say a process inside the sys container is in a dedicated mount namespace has a root of /var/lib/; if that process decides to unmount /docker, it's really unmounting /var/lib/docker inside the sys container, so it should not be allowed (since /var/lib/docker is an immutable mount). But this code will allow it.

seccomp/umount.go
rodnymolina: I'll rewrite it to incorporate some of this, but keep in mind that we are always talking about an umount instruction generated form within an unshared() context; all other cases that I can think of involve inner containers, and for those, a chroot/pivot_root instruction is involved, which prevent this corner-case scenario from happening.

seccomp/umount.go
```diff
@@ -117,15 +124,81 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+// unmounted.
+//
+// Method will return 'true' if the umount operation is deemed legit, and will
+// return 'false' otherwise.
+func (u *umountSyscallInfo) processUmountAllowed(
+	mip domain.MountInfoParserIface) (bool, *sysResponse) {
+
+	// There must be mountinfo state present for this target. Otherwise, let
+	// kernel handle the error.
+	info := mip.GetInfo(u.Target)
```
ctalledo, line 137: The `u.Target` is the target (absolute) path with respect to the process' root (which may not be necessarily the sys container's root). 

It seems to me we have to convert that to the target with respect to the sys container's root.  The reason is that the mountInfoParser has the paths with respect to the sys container's root (correct?). 

If we don't adjust `u.Target` accordingly, then the code will get confused. For example, a process that chroots to `/var/lib` and then unmounts `/docker` will be allowed to unmount. That's because `u.Target = /docker`, so `mip.GetInfo()` will return `nil`.

If we do adjust `u.Target` as I am suggesting, then it seems to me that the `if p.RootInode() != u.cntr.InitProc().RootInode()` check below is no longer needed.

ctalledo: The problem is that we don't have a way of adjusting u.Target relative to the sys container's root, because the kernel does not provide the info that will allow us to do this.

So what we need to do instead, it seems to me, is to have the mountInfoParser store the inode of each immutable mount. This way, when u.Target comes in, we can get it's inode (via `stat()`), and compare it unequivocally with the inodes for the immutable mounts. This works regardless of whether the process doing the unmount chroote'd or not. 
It also allows us to not care about whether the process unshared or not it's namespace.

ctalledo: Ok got it, my mistake.

seccomp/umount.go
ctalledo: Got it.

seccomp/umount.go
ctalledo: The bug is present. It occurs when a process unshares its mount ns, does a chroot to `/var/lib/` and unmounts `/docker`.

In this case the `IsImmutableMountID()` check above won't catch it (due to the new mount namespace) and since the process did a chroot, this code will allow the umount to succeed, when it shouldn't have.

ctalledo: It seems to me you want to first check if the process doing the umount is in the same mount namespace as the sys container's root process. 

If it is, then the IsImmutableMountID() check suffices. 

If it is not, then we can't check by mount-ID, so we have to check by other means. Currently the code checks by umount target path, but in order to do that you have to know the path with respect to the sys container's root. I don't see the code doing that. In fact we can't do that as I mentioned in another comment, as we don't have enough info from the kernel to convert the umount target path to a path relative to the container's root. The only thing that comes to mind is that we check by umount target path inode. 

In addition, we have to check if there is a mount on top of the immutable mount. If there is, the umount should be allowed to proceed. Function `IsRecursiveBindMount` attempts to do that, but it's not catching non-bind mounts (e.g., a tmpfs mount on top of an immutable mount for example).

mount/infoParser.go
```diff
@@ -0,0 +1,1037 @@
...
+
+		if elem.Root == info.Root && elem.Source == info.Source {
+			return mh.StringToFlags(elem.Options)&unix.MS_RDONLY == unix.MS_RDONLY
+		}
+	}
+
+	return false
+}
+
+func (mi *mountInfoParser) equivalentMounts(m1, m2 *domain.MountInfo) bool {
```
rodnymolina, line 859: Disregard this function (equivalentMounts) and the next one (equivalentMountAncestors) as I didn't use them at the end. Will remove them.

nsenter/event.go
```diff
@@ -471,6 +551,37 @@ func (e *NSenterEvent) ReceiveResponse() *domain.NSenterMessage {
 	return e.ResMsg
 }
 
+// TerminateRequest serves to unwind the nsenter-event FSM after the generation
+// of an asynchronous event through SendRequest(true) execution.
+func (e *NSenterEvent) TerminateRequest() error {
```
rodnymolina, line 556: Need to revisit this comment.

mount/infoParser.go
```diff
@@ -0,0 +1,1037 @@
...
+	return info.Root == infoParent.Root &&
+		info.MountPoint == infoParent.MountPoint &&
+		info.Source == infoParent.Source
+}
+
+// IsOverlapMount determines if the mountpoint associated to a mountInfo entry
+// overlaps with any other mountpoint in the mountInfo tree. A 'self' mount
+// is a special case (subset) of an 'overlap' one.
+//
+// Example 1: Same as as the above one (IsSelfMount method).
```
ctalledo, line 781: typo "as as"

mount/infoParser.go
```diff
@@ -0,0 +1,1037 @@
...
+//
+// 2712 2192 0:153 / /usr/src/linux-headers-5.4.0-62 ro,relatime - shiftfs /usr/src/linux-headers-5.4.0-62 rw
+// 3074 2712 0:153 / /usr/src/linux-headers-5.4.0-62 ro,relatime - shiftfs /usr/src/linux-headers-5.4.0-62 rw
+//
+// Example 2: mountID-3074 is an 'overlap' mount of the original mountID-2706 entry ]
```
ctalledo, line 786: Typo: remo "]" at the end.

mount/infoParser.go
```diff
@@ -0,0 +1,1037 @@
...
+	p2 := m2.Mip.GetParentMount(m2)
+	if p2 != nil && m2.Mip.IsSysboxfsBaseMount(p2.MountPoint) {
+		p2 = m2.Mip.GetParentMount(p2)
+	}
+
+	return mi.equivalentMounts(p1, p2)
+}
+
+// IsCloneMount determines if the passed mountInfo entry is a 'clone' of an
+// entry in the sys-container's mount namespace.
```
ctalledo, line 895: There is no notion of "sys container" at this level of the code, so let's rephrase.

ctalledo: Suggestion: add the following:
"It does this by comparing the mountpoint attributes of the given mount with those of the mi object. If the mountpoint attributes are not sufficient, it also compares the attributes of the parent/ancestor mounts."

mount/infoParser.go
```diff
@@ -0,0 +1,1037 @@
...
+		return false
+	}
+
+	var candidateList []*domain.MountInfo
+
+	// Extract the list of mountpoints matching the incoming process' maj-min-id.
+	devIdSlice := mi.devIdInfo[procInfo.MajorMinorVer]
+
+	for _, cntrInfo := range devIdSlice {
+		if cntrInfo.MountID == procInfo.MountID {
```
ctalledo, line 911: Suggestion: add this comment here:
"// A mountpoint with the same ID can't be a clone (by definition)"

mount/infoParser.go
```diff
@@ -0,0 +1,1037 @@
...
+
+		// Skip readwrite candidates when operating in readonly mode.
+		if ronlyMatch && !mi.IsRoMount(cntrInfo) {
+			continue
+		}
+
+		// All candidates must meet a minimum set of criteria.
+		if cntrInfo.Root != procInfo.Root ||
+			cntrInfo.Source != procInfo.Source ||
+			mi.IsRoMount(cntrInfo) != mi.IsRoMount(procInfo) {
```
ctalledo, line 923: We should check the mismatch of all flags, except the read-only flag. That's because a clone mount must match on all attributes (except read-only since we allow read-only remounting of immutable mounts).

mount/infoParser.go
```diff
@@ -0,0 +1,1037 @@
...
+		if m2.MpInode == 0 {
+			err := m2.Mip.ExtractAncestorInodes(m2)
+			if err != nil {
+				return false
+			}
+		}
+
+		// Return 'false' whenever a mismatch is found in any of the elements
+		// of the ancestry line.
+		if m1.MpInode != m2.MpInode {
```
ctalledo, line 1002: Let's compare the ancestor root, source, and flags (i.e., "the minimum set of criteria" we compared in IsCloneMount()).

seccomp/umount.go
```diff
@@ -113,19 +120,237 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	//    | 8         | yes          | yes          | yes      |
+	//    +-----------+--------------+--------------+----------+
+	//
+
+	// Identify the mount-ns of the process launching the unmount to compare it
+	// with the one of the sys container's initpid. In the unlikely case of an
+	// error, let the kernel deal with it.
+	processMountNs, err := u.processInfo.MountNsInode()
+	if err != nil {
+		return false, u.tracer.createContinueResponse(u.reqId)
```
ctalledo, line 181: We should explicitly return an error here; we don't want to rely on the kernel doing this (what if it allows the unmount, then we would have a security hole).

seccomp/umount.go
```diff
@@ -113,19 +120,237 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	}
+	initProcMountNs, err := u.cntr.InitProc().MountNsInode()
+	if err != nil {
+		return false, u.tracer.createContinueResponse(u.reqId)
```
ctalledo, line 185: Same as prior comment.

seccomp/umount.go
```diff
@@ -113,19 +120,237 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	// rely on the target's mountID to discern an immutable target from a
+	// regular one. Otherwise, we cannot rely on the mountID field, as the values
+	// allocated by kernel for these very mountpoints will differ in other mount
+	// namespaces.
+	if processMountNs == initProcMountNs {
+
+		if u.processInfo.Root() == "/" {
+			processRootInode := u.processInfo.RootInode()
+
+			// Scenario 1): no-unshare(mnt) & no-pivot() & no-chroot()
```
ctalledo, line 201: Scenarios 1->4 below result result in the same action: checking if the mount is immutable by ID. Thus, I suggest not differentiating between them and instead merging them into a single call to "IsImmutableMountID". In fact, the call to mip.ExtractInode("/") in scenarios 3 and 4 is expensive (due to the nsenter) and not really necessary.

This will simplify the code: if the process doing the unmount is in the same mount ns as the sys container init process, we simply lookup the mount by ID and we are done.

seccomp/umount.go
```diff
@@ -113,19 +120,237 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+		}
+
+	} else {
+
+		if u.processInfo.Root() == "/" {
+			processRootInode := u.processInfo.RootInode()
+
+			// Scenario 5): unshare(mnt) & no-pivot() & no-chroot()
+			if processRootInode == syscntrRootInode {
+				if mip.IsOverlapMount(info) {
```
ctalledo, line 263: This assumes that an immutable mount can't possibly be an overlap mount (i.e., two or more mounts stacked over the same mountpoint). That's a risky assumption. Is there a more robust way of knowing if the overlap mount is in fact immutable or not?

state/container.go
```diff
@@ -140,6 +160,74 @@ func (c *container) String() string {
 		c.id, int(c.initPid), c.uidFirst, c.gidFirst)
 }
 
+func (c *container) IsImmutableMountID(id int) bool {
+	c.intLock.RLock()
+	defer c.intLock.RUnlock()
+
+	if info := c.mountInfoParser.LookupByMountID(id); info != nil {
```
ctalledo, line 167: Suggestion: add comment:
"All mounts initially present in the container are considered immutable."

mount/infoParser.go
```diff
@@ -0,0 +1,1037 @@
...
+		// compare in either ancestry line.
+		if m1 == nil || m2 == nil {
+			return true
+		}
+
+		// We must skip basemount components of sysbox-fs' managed file-systems
+		// (i.e. /proc & /sys), due to the particular hierarchy displayed by
+		// their submounts as a consequence of the specific approach sysbox-fs
+		// follows to bind-mount these resources during container initialization.
+		if mi.isSysboxfsBaseMount(m1) {
```
ctalledo, line 980: Not clear to me why we need this here ...

seccomp/umount.go
```diff
@@ -113,19 +120,221 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	if u.FsType == "proc" || u.FsType == "sysfs" {
+		return true, nil
+	}
+
+	// There must be mountinfo state present for this target. Otherwise, let
+	// kernel handle the error.
+	info := mip.GetInfo(u.Target)
+	if info == nil {
+		return false, u.tracer.createContinueResponse(u.reqId)
+	}
```
ctalledo, line 151: Should we return error here, instead of letting the kernel handle it?

seccomp/mount.go
```diff
@@ -523,22 +559,279 @@ func (m *mountSyscallInfo) createNfsMountPayload(
...
+	// Approve operation if it attempts to remount target as read-only.
+	if mh.IsReadOnlyMount(m.Flags) {
+		return true, nil
+	}
+
+	// There must be mountinfo state present for this target. Otherwise, let
+	// kernel handle the error.
+	info := mip.GetInfo(m.Target)
+	if info == nil {
+		return false, m.tracer.createContinueResponse(m.reqId)
```
ctalledo, line 586: Should we return error here instead of letting the kernel handle it?

seccomp/mount.go
```diff
@@ -523,22 +559,279 @@ func (m *mountSyscallInfo) createNfsMountPayload(
...
+			processRootInode := m.processInfo.RootInode()
+
+			// Scenario 5): unshare(mnt) & no-pivot() & no-chroot()
+			if processRootInode == syscntrRootInode {
+
+				// In this scenario we have full access to all the mountpoints
+				// within the sys-container (different mount-id though), so we
+				// can safely rely on the mountpoints to determine resource's
+				// immutability.
+				if m.cntr.IsImmutableMountpoint(info.MountPoint) {
```
ctalledo, line 736: This check is not sufficient, as it won't detect overlapping mounts.

seccomp/mount.go
```diff
@@ -523,22 +559,279 @@ func (m *mountSyscallInfo) createNfsMountPayload(
...
+			}
+
+			// Scenario 7): unshare(mnt) & no-pivot() & chroot()
+			if processRootInode == syscntrRootInode {
+
+				// In this scenario we have full access to all the mountpoints
+				// within the sys-container (different mount-id though), so we
+				// can safely rely on the mountpoints to determine resource's
+				// immutability.
+				if m.cntr.IsImmutableMountpoint(info.MountPoint) {
```
ctalledo, line 785: Same as prior comment: this does not detect overlapping mounts.

ctalledo (commented):
A few more comments as we discussed yesterday.

seccomp/umount.go
```diff
@@ -113,19 +120,221 @@ func (u *umountSyscallInfo) process() (*sysResponse, error) {
...
+	syscntrRootInode := u.cntr.InitProc().RootInode()
+
+	// If process' mount-ns matches the sys-container's one, then we can simply
+	// rely on the target's mountID to discern an immutable target from a
+	// regular one. Otherwise, we cannot rely on the mountID field, as the values
+	// allocated by kernel for these very mountpoints will differ in other mount
+	// namespaces.
+	if processMountNs == initProcMountNs {
+
+		if ok := u.cntr.IsImmutableMountID(info.MountID); ok == true {
```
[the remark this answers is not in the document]
rodnymolina: Sure.

rodnymolina (commented):
Latest patch uploaded.

seccomp/umount.go
ctalledo: It's better for us to return the error, otherwise we open a TOCTOU error: by the time the kernel checks, who's to say that the mountpoint won't be there?

ctalledo (approved):
With the latest changes it looks good, great job!

Let's only add the sysbox-fs config options we discussed to control the behavior of umount and remount (e.g., `initial-mount-allow-umount` and `initial-mount-allow-remount`. We will probably need a couple of tests for it too.

rodnymolina (commented):
Done

PR #25, opened 2021-01-25
Initial fix to enable sysbox containers with ArchLinux + systemd

ctalledo:


ctalledo:
FYI, all Sysbox integration and unit tests pass with this change.

nsenter/event.go
```diff
@@ -767,9 +824,48 @@ func (e *NSenterEvent) processChownSyscallRequest() error {
 	return nil
 }
 
+func (e *NSenterEvent) getProcCreds(pipe *os.File) error {
```
rodnymolina, line 827: Why aren't you using the routines that we have in sysbox-libs for this purpose (i.e. `io.go`) by extending them if needed? Those routines were built making use of the proper abstractions provided by golang's unix pkg, so that we avoid calling syscalls directly. Also, keep in mind that there's already code in sysbox-fs making use of these routines, so we would end up with overlapping logic within the same daemon.

nsenter/event.go
```diff
@@ -629,22 +657,22 @@ func (e *NSenterEvent) processMountSyscallRequest() error {
...
-		// Create a 'process' struct to represent the 'sysbox-fs nsenter' process
-		// executing this logic.
-		process := e.service.prs.ProcessCreate(0, 0, 0)
 
-		// Extract payload-header from the first element (just one in overlayfs
-		// mounts) of the payload slice .
-		header := payload[0].Header
+		// Create a dummy 'process' struct to represent the 'sysbox-fs nsenter' process
+		// executing this logic.
+		this := e.service.prs.ProcessCreate(0, 0, 0)
```
rodnymolina, line 671: Why using 'this' as a variable name instead of 'process' as you do below? Also, in my mind `this` carries special  connotations (from java/c++) that don't seem to apply in this context.

rodnymolina (commented):
Not fully done yet, have a couple of comments so far.

nsenter/event.go
ctalledo: I wanted to differentiate between two processes: the sysbox-fs nsenter process ("this"), and the process on whose behalf we are doing some action ("process").  I liked the name "this" because if fits perfectly in "this.AdjustPersonality()".

nsenter/event.go
ctalledo: Oh I had not noticed; let me see if I can move the code into `io.go` then. You are right that it's better to use `unix` instead of `syscall`.

nsenter/event.go
```diff
@@ -404,7 +410,29 @@ func (e *NSenterEvent) SendRequest() error {
...
 	cmd.Process = process
 
+	//
 	// Transfer the nsenterEvent details to grand-child for processing.
+	//
+
+	// Send the pid using SCM rights, so it shows up properly inside the
+	// nsexec process.
+	//
+	// TODO: in the future we should also send the process uid and gid
```
rodnymolina, line 420: We are already passing uid/gid params in a few syscalls by directly including this info into the nsevent's payload (e.g. overlayfs mounts). Obviously, in those cases we're not entering the backend's user-ns, that's why it works fine. Now, I agree that this approach is more generic/consistent, but you may want to update this comment to reflect that.

nsenter/event.go
rodnymolina: Fine.

nsenter/event.go
ctalledo: I tried hard to do this, but it's a pain unfortunately. The ipc stuff we have on io.go works using *net.UnixConn, and it turns out that setting SO_PASSCRED on the socket (needed to pass credentials across it) is a pain because there is no straight forward mechanism to do it (we can't use unix.SetSockoptInt() as I was hoping). I'll look into it a bit more, but don't want to spend much more time on this.

rodnymolina (approved):
We agreed to merge this one as is, and do the refactorization of the credentials-passing logic right after to make use of higher-level api's.

nsenter/event.go
ctalledo: I updated the comment, thanks!

nsenter/event.go
ctalledo: Rodny and I agreed to do this in a subsequent refactoring.

PR #27, opened 2021-02-06
Fix deadlock in sysbox-fs seccomp tracer.

ctalledo:


seccomp/tracer.go
```diff
@@ -171,59 +189,62 @@ func (t *syscallTracer) start() error {
...
+
+	s, ok := t.seccompSessionMap[pid]
+	if !ok {
+		logrus.Errorf("Unexpected error: seccomp fd not found for pid %d", pid)
+		t.seccompSessionMu.Unlock()
+		return
+	}
+
+	if err := syscall.Close(int(s.fd)); err != nil {
+		logrus.Fatal(err)
```
rodnymolina, line 222: I know that i was the one that added this 'fatal' instruction here, but i'm specially concerned given that you are considerably expanding the pidMonitor interval, so there's more room now for Close() to return error if kernel (and other goroutines) has fully cleaned all process' resources. My point: can you substitute this 'fatal' log by a 'warning'?

state/container_test.go
```diff
@@ -167,38 +167,6 @@ func Test_container_Data(t *testing.T) {
 	}
 }
 
-func Test_container_String(t *testing.T) {
```
rodnymolina, line 170: Minor: why are you removing this? This method (mount stringer) is used in various places which may be expecting a specific output format. Definitely not a critical UT, but it doesn't bother us either.

rodnymolina (changes requested):
Looks good. A couple of comments though.

seccomp/tracer.go
rodnymolina: Actually, now that i think about it, this Close() syscall shouldn't be needed at all as the process should be dead by now ...

state/container_test.go
ctalledo: I removed it because the test was failing (due to a change I did earlier in sysbox-fs logging), and looking at the function, I saw it was not worth keeping/maintaining since it only tests the stringer function, which simply logs something.

seccomp/tracer.go
ctalledo: I disagree. This is closing an fd owned by sysbox-fs, even though the fd tracks seccomp on a process. Even if the traced process dies, the fd still needs to be closed (otherwise sysbox-fs will have an file-open leak). And closing the fd should not return an error, because it's simply closing a sysbox-fs owned fd; thus, I think it's fine to log a strong error when this occurs.

Am I missing something here?

rodnymolina: You are right that sysbox-fs must close their side of the seccomp pipe (fd). But I had some doubts about this  'logrus.Fatal' instruction here as it would bring the entire sysbox-fs down (panic). On the other hand, I now realize that this method (removeSession) runs within the scope of sessionsMonitor common thread, so we wouldn't be reducing the failure-scope in any way by removing this 'panic'.

rodnymolina (approved):
LG

seccomp/tracer.go
ctalledo: I went ahead and modified the logrus.Fatal() -> logrus.Errorf(), as the former will stop the sysbox-fs process.

PR #28, opened 2021-02-11
Improve the cap_last_cap handler to reduce number of open file descriptors.

ctalledo:
Prior to this change, the sysbox-fs handler for /proc/sys/kernel/cap_last_cap
would open a file descriptor whenever a process in a container opened this
resource. And it would keep it open until the process in the container closed
the resource. Given that cap_last_cap is one of the most (if not the most)
accessed sysctl by processes inside containers, this led to sysbox-fs having a
very large number of file descriptors opened at a given time.

This change improves on this, by having sysbox-fs only open the cap_last_cap
handler when it's value is read, and closes it immediately after.  This
significantly reduces the number of open fds that sysbox-fs has at a any given
time.

### DIFF
--- a/handler/implementations/kernelLastCap.go
+++ b/handler/implementations/kernelLastCap.go
@@ -36,6 +36,10 @@ import (
 // highest capability supported by the running kernel ('37' as of today's
 // latest / 5.X kernels ).
 //
+// This handler is used for performance reasons (rather than functional reasons),
+// as having it avoids using the /proc/sys common handler for accesses to
+// /proc/sys/kernel/cap_last_cap which is the most commonly accessed sysctl.
+//
 type KernelLastCapHandler struct {
 	domain.HandlerBase
 }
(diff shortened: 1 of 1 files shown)


handler/implementations/kernelLastCap.go
```diff
@@ -69,23 +73,13 @@ func (h *KernelLastCapHandler) Open(
 		return fuse.IOerror{Code: syscall.EACCES}
 	}
```
rodnymolina, line 75: As this deviates from our usual approach of opening/closing the fds across all our handlers, please add a note here to explain why we are allowing it in this particular case. Something like ...

"In this particular handler we are not opening the FS node as there is no need to validate the process credentials to access this resource (i.e. /proc/sys/kernel/cap_last_cap is read-only for all processes in the system). This is done for optimization purposes to save cycles ... blah blah ..."

rodnymolina (approved):
See minor comment.

handler/implementations/kernelLastCap.go
ctalledo: Talked with Rodny and we agreed that what the cap_last_cap handler is doing should be the default action for most handlers, so I'll skip the comment.

PR #30, opened 2021-02-27
Fix for bogus log generated during container registration

rodnymolina:
This one is a side effect of a change made a few days ago to fix a potential deadlock situation in container.String() method. At that time i decided to make this method an internal one of the container class, but didn't notice that there was one external package making use of this method.

As a consequence of the above, the following bogus log is now being displayed during container registration:

```
time="2021-02-27 06:18:19" level=info msg="Container registration completed: &{{{0 0} 0 0 0 0} 03eefd5f9d3ea3717727d82c42ba876360529841557ce8a8051d13da191f7d2c 8763 0 {0 0 <nil>} 165536 65536 165536 65536 [/proc/bus /proc/fs /proc/irq /proc/sysrq-trigger] [/proc/asound /proc/acpi /proc/keys /proc/latency_stats /proc/timer_list /proc/timer_stats /proc/sched_debug /proc/scsi] <nil> map[] <nil> 0xc00025a230 {{0 0} 0 0 0 0} {0
```

A proper fix for this one is to avoid the utilization of the container.string() method by external packages, especially when the information being logged is related to properties (registrations / unregistrations) that are owned by the container package itself; this state and associated logs belongs within the container/containerDB realm.

### DIFF
--- a/ipc/apis.go
+++ b/ipc/apis.go
@@ -65,8 +65,6 @@ func (ips *ipcService) Init() error {
 
 func ContainerPreRegister(ctx interface{}, data *grpc.ContainerData) error {
 
-	logrus.Debugf("Container pre-registration started: id = %s", data.Id)
-
 	ipcService := ctx.(*ipcService)
 
 	err := ipcService.css.ContainerPreRegister(data.Id)
@@ -74,14 +72,10 @@ func ContainerPreRegister(ctx interface{}, data *grpc.ContainerData) error {
 		return err
 	}
 
-	logrus.Debugf("Container pre-registration completed: id = %s", data.Id)
-
 	return nil
 }
 
 func ContainerRegister(ctx interface{}, data *grpc.ContainerData) error {
-
-	logrus.Debugf("Container registration started: id = %s", data.Id)
 
 	ipcService := ctx.(*ipcService)
 
@@ -105,14 +99,10 @@ func ContainerRegister(ctx interface{}, data *grpc.ContainerData) error {
 		return err
 	}
 
-	logrus.Infof("Container registration completed: %v", cntr)
-
 	return nil
 }
 
 func ContainerUnregister(ctx interface{}, data *grpc.ContainerData) error {
-
-	logrus.Debugf("Container unregistration started: id = %s", data.Id)
 
 	ipcService := ctx.(*ipcService)
 
@@ -131,14 +121,10 @@ func ContainerUnregister(ctx interface{}, data *grpc.ContainerData) error {
 		return err
 	}
 
-	logrus.Infof("Container unregistration completed: id = %s", data.Id)
-
 	return nil
 }
 
 func ContainerUpdate(ctx interface{}, data *grpc.ContainerData) error {
-
-	logrus.Debugf("Container update started: id = %s", data.Id)
 
 	ipcService := ctx.(*ipcService)
 
@@ -161,8 +147,6 @@ func ContainerUpdate(ctx interface{}, data *grpc.ContainerData) error {
 	if err != nil {
 		return err
 	}
-
-	logrus.Debugf("Container update completed: id = %s", data.Id)
 
 	return nil
 }
--- a/state/containerDB.go
+++ b/state/containerDB.go
@@ -101,6 +101,9 @@ func (css *containerStateService) ContainerCreate(
 }
 
 func (css *containerStateService) ContainerPreRegister(id string) error {
+
+	logrus.Debugf("Container pre-registration started: id = %s", id)
+
 	css.Lock()
 
 	// Ensure that new container's id is not already present.
@@ -136,13 +139,18 @@ func (css *containerStateService) ContainerPreRegister(id string) error {
 
 	css.Unlock()
 
+	logrus.Debugf("Container pre-registration completed: id = %s", id)
+
 	return nil
 }
 
 func (css *containerStateService) ContainerRegister(c domain.ContainerIface) error {
-	css.Lock()
 
 	cntr := c.(*container)
+
+	logrus.Debugf("Container registration started: id = %s", cntr.id)
+
+	css.Lock()
 
 	// Ensure that container's id is already present (pregistration completed).
 	currCntr, ok := css.idTable[cntr.id]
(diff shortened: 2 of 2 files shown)


ctalledo:
Often Golang objects provide a String() method to satisfy the stringer interface, such that entities that use the object can print the object's properties. Thus I am not sure why String() needs to be made an internal method of the container object.

rodnymolina:
> Often Golang objects provide a String() method to satisfy the stringer interface, such that entities that use the object can print the object's properties. Thus I am not sure why String() needs to be made an internal method of the container object.

I agree with you on this part. Ideally, opaque types (specially large objects) should satisfy the stringer interface. Unfortunately, after several hours debugging this issue (in essence, a `testify` limitation), I decided to go for a workaround that simply hides the String() method -- not a big deal at this point given that there's hardly no users for this method.

Now, what's the issue? It turns out that as part of its mockery magic, `testify` invokes the stringer methods of the objects being mocked. In our case, that translates into a deadlock situation coz, at container registration, we acquire a lock on the container on which we are operating, but then the method being mocked (NewMountInfoParser()) also attempt to obtain a lock on this same container. See stack-trace below.

Problem has been [reported](https://github.com/stretchr/testify/issues/643) a few times for there's no workaround/solution for it yet. And unfortunately, I don't see any obvious adjustment to the testcase to allow us to bypass this limitation.

IMO, given how critical this code-path is, if we need to compromise between having this testcase and having a public String() method, I would rather keep the testcase, specially since there are no consumers (only one before my changes) of this String() method.

```console
goroutine 15 [semacquire]:
sync.runtime_SemacquireMutex(0xc00010660c, 0xc0000bf500, 0x0)
	/usr/local/go/src/runtime/sema.go:71 +0x47
sync.(*RWMutex).RLock(...)
	/usr/local/go/src/sync/rwmutex.go:50
github.com/nestybox/sysbox-fs/state.(*container).String(0xc000106540, 0x0, 0x0)
	/home/rmolina/wsp/01-11-2021/sysbox/sysbox-fs/state/container.go:322 +0x233
fmt.(*pp).handleMethods(0xc00008edd0, 0x76, 0x95e901)
	/usr/local/go/src/fmt/print.go:630 +0x302
fmt.(*pp).printArg(0xc00008edd0, 0x94b7c0, 0xc000106540, 0x76)
	/usr/local/go/src/fmt/print.go:713 +0x206
fmt.(*pp).doPrintf(0xc00008edd0, 0x95e98c, 0xd, 0xc0000bd830, 0x1, 0x1)
	/usr/local/go/src/fmt/print.go:1126 +0x406
fmt.Sprintf(0x95e98c, 0xd, 0xc0000bd830, 0x1, 0x1, 0x4, 0xd4)
	/usr/local/go/src/fmt/print.go:219 +0x66
github.com/stretchr/testify/mock.Arguments.Diff(0xc0000aca50, 0x5, 0x5, 0xc0000acaa0, 0x5, 0x5, 0x28, 0x9048a0, 0xd4c0c0)
	/home/rmolina/go/pkg/mod/github.com/stretchr/testify@v1.6.1/mock/mock.go:768 +0x1297
github.com/stretchr/testify/mock.(*Mock).findExpectedCall(0xc0000ac4b0, 0xcdd521, 0x12, 0xc0000acaa0, 0x5, 0x5, 0x8, 0x4ee161)
	/home/rmolina/go/pkg/mod/github.com/stretchr/testify@v1.6.1/mock/mock.go:287 +0xff
github.com/stretchr/testify/mock.(*Mock).MethodCalled(0xc0000ac4b0, 0xcdd521, 0x12, 0xc0000acaa0, 0x5, 0x5, 0xc0000215c0, 0x4, 0x4)
	/home/rmolina/go/pkg/mod/github.com/stretchr/testify@v1.6.1/mock/mock.go:366 +0xa7
github.com/stretchr/testify/mock.(*Mock).Called(0xc0000ac4b0, 0xc0000acaa0, 0x5, 0x5, 0xc000054d38, 0x40e646, 0xc0000100f8)
	/home/rmolina/go/pkg/mod/github.com/stretchr/testify@v1.6.1/mock/mock.go:356 +0x1a7
github.com/nestybox/sysbox-fs/mocks.(*MountServiceIface).NewMountInfoParser(0xc0000ac4b0, 0xa0e720, 0xc000106540, 0xa0e180, 0xc0000b5ae0, 0xc000010101, 0xd4c0c0, 0x7f4b9d46f008, 0x0, 0x203000)
	/home/rmolina/wsp/01-11-2021/sysbox/sysbox-fs/mocks/MountServiceIface.go:49 +0x156
github.com/nestybox/sysbox-fs/state.(*container).update(0xc000106540, 0xc000106540, 0x0, 0x0)
	/home/rmolina/wsp/01-11-2021/sysbox/sysbox-fs/state/container.go:293 +0x302
github.com/nestybox/sysbox-fs/state.(*containerStateService).ContainerRegister(0xc0000f3340, 0xa0e720, 0xc000106540, 0x26d899e6bc589, 0x37f83c63)
	/home/rmolina/wsp/01-11-2021/sysbox/sysbox-fs/state/containerDB.go:169 +0x157
github.com/nestybox/sysbox-fs/state.Test_containerStateService_ContainerRegister.func5(0xc0000fc900)
	/home/rmolina/wsp/01-11-2021/sysbox/sysbox-fs/state/containerDB_test.go:481 +0x138
testing.tRunner(0xc0000fc900, 0xc000057830)
	/usr/local/go/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:960 +0x350
FAIL	github.com/nestybox/sysbox-fs/state	600.014s
FAIL
rmolina@dev-vm1:~/wsp/01-11-2021/sysbox/sysbox-fs$
```
 

> In fact it's cleaner if the call to logrus are outside the object itself, so that the caller can decide when to log or not log info about the object.
> 
> I understand the info being logged is related to properties owned by the object, but that's the reason for the String() method (it's implementation is controlled by the object itself). This does not mean that the object should control when to log info about itself; that should be up to the caller.

This is the part I don't quite agree with. It depends. In some cases you want your class to fully own and control the operation of its objects, including UI/UX aspects. And if there's a client that requires special handling, the class should then export a parameter to offer the client a way to control how that operation is performed (e.g. log-on/off).

ctalledo (approved):

PR #34, opened 2021-04-12
fix: Use os.Stdout if no logging filename is given

nudgegoonies:
* Use os.Stderr if no logging filename is given
* Set default log to "" to enable stderr output
* This makes it possible to use journald by using default log value
* This solution is not optimal, but keeps backwards compatibility

Refs: https://github.com/nestybox/sysbox/issues/221

### DIFF
--- a/cmd/sysbox-fs/main.go
+++ b/cmd/sysbox-fs/main.go
@@ -187,8 +187,8 @@ func main() {
 		},
 		cli.StringFlag{
 			Name:  "log",
-			Value: "/dev/stdout",
-			Usage: "log file path",
+			Value: "",
+			Usage: "log file path or empty string for stderr output (default: \"\")",
 		},
 		cli.StringFlag{
 			Name:  "log-level",
@@ -264,6 +264,9 @@ func main() {
 
 			logrus.SetOutput(f)
 			log.SetOutput(f)
+		} else {
+			logrus.SetOutput(os.Stderr)
+			log.SetOutput(os.Stderr)
 		}
 
 		if logFormat := ctx.GlobalString("log-format"); logFormat == "json" {


cmd/sysbox-fs/main.go
```diff
@@ -264,6 +264,9 @@ func main() {
 
 			logrus.SetOutput(f)
 			log.SetOutput(f)
+		} else {
+			logrus.SetOutput(os.Stdout)
```
ctalledo, line 268: Same as for the sysbox-mgr PR: let's log to `os.stderr` instead.

ctalledo (changes requested):
Thanks for the PR!

rodnymolina (changes requested):
Not sure why but i'm seeing this in my setup. @nudgegoonies, can you reproduce this too?

Let's sort this out first.

```
$ docker run --runtime=sysbox-runc -it --rm --name=test-1 --hostname=test-1 ubuntu
docker: Error response from daemon: OCI runtime create failed: container_linux.go:392: starting container process caused: process_linux.go:592: container init caused: write sysctl key net.ipv4.ip_unprivileged_port_start: open /proc/sys/net/ipv4/ip_unprivileged_port_start: no such file or directory: unknown.
$
```

cmd/sysbox-fs/main.go
```diff
@@ -187,7 +187,7 @@ func main() {
 		},
 		cli.StringFlag{
 			Name:  "log",
-			Value: "/dev/stdout",
+			Value: "/var/log/sysbox-fs.log",
 			Usage: "log file path",
```
mfriedenhagen, line 191: The usage hint for logging should probably include the hint that using "" will redirect to os.Stderr

cmd/sysbox-fs/main.go
nudgegoonies: Done!

nudgegoonies:
> Not sure why but i'm seeing this in my setup. @nudgegoonies, can you reproduce this too?
> 
> Let's sort this out first.
> 
> ```
> $ docker run --runtime=sysbox-runc -it --rm --name=test-1 --hostname=test-1 ubuntu
> docker: Error response from daemon: OCI runtime create failed: container_linux.go:392: starting container process caused: process_linux.go:592: container init caused: write sysctl key net.ipv4.ip_unprivileged_port_start: open /proc/sys/net/ipv4/ip_unprivileged_port_start: no such file or directory: unknown.
> $
> ```
No, this command works perfect on my Debian 10 Server:
```
docker run --runtime=sysbox-runc -it --rm --name=test-1 --hostname=test-1 ubuntu
root@test-1:/# 

```

cmd/sysbox-fs/main.go
nudgegoonies: You are right. But to be true, this output is for people starting sysbox-fs and sysbox-mgr manually from console. I assume those people expect Stderr to be the default and not /var/log/sysbox-fs/mgr.log. So the default should be "". But this looks strange when people run -h or --help because the Value is printed out there and also used as default value. And before my fix the Value was the non working /dev/stdout so i think my patch was not so backward compatible then. But i think we can change it to "" while adding documentation. And the systemd sets this option and is independent of the default value. I try.

rodnymolina:
Thanks @nudgegoonies, let me see what's going on my side then ...

nudgegoonies:
> Thanks @nudgegoonies, let me see what's going on my side then ...

@rodnymolina I forgot to tell, that i was using the 0.3.0. Now i tried a fresh master build and it works too.

ctalledo (approved):
Looks good, thanks!

Please sign off the commit (i.e., git commit --sign-off ...) as we need this in order to accept the patch. See [here](https://github.com/nestybox/sysbox/blob/master/CONTRIBUTING.md#sign-your-work) for more.

nudgegoonies:
> Looks good, thanks!
> 
> Please sign off the commit (i.e., git commit --sign-off ...) as we need this in order to accept the patch. See [here](https://github.com/nestybox/sysbox/blob/master/CONTRIBUTING.md#sign-your-work) for more.

Done. Don't be confused by the email change. I forked the project into my private namespace and not into the company namespace.

rodnymolina (approved):
LG @nudgegoonies!

Not sure what went wrong on my side but can't repro anymore.

PR #36, opened 2021-05-15
Fix deadlock issue around reaper's mutex during async-nsenter execution

rodnymolina:
Problem diagnosis has been well documented as part of issue [#266](https://github.com/nestybox/sysbox/issues/266). Here I'm briefly explaining the fix that I've identified for this bug.

As described in the issue's notes, the problem is caused by a deadlock scenario while dealing with the reaper's RWmutex. The fix is fairly simple: have 'async' nsenter logic remove its reaper's dependency. See that there's no real need for the reaper in the 'async' nsenter case as the SendRequest() callee will always sigkill() the nsenter process, so we can safely remove all this concurrency complexity in this case.

Fix has been heavily tested over the last couple of weeks in the same scaling scenario where the problem was originally reported. No deadlock issues has been observed ever since.

### DIFF
--- a/domain/nsenter.go
+++ b/domain/nsenter.go
@@ -132,8 +132,6 @@ type NSenterEventIface interface {
 	SetResponseMsg(m *NSenterMessage)
 	GetResponseMsg() *NSenterMessage
 	GetProcessID() uint32
-	ReapStartRequest()
-	ReapEndRequest()
 }
 
 // NSenterMessage struct defines the layout of the messages being exchanged
--- a/domain/nsenter.go
+++ b/domain/nsenter.go
@@ -132,6 +132,8 @@ type NSenterEventIface interface {
 	SetResponseMsg(m *NSenterMessage)
 	GetResponseMsg() *NSenterMessage
 	GetProcessID() uint32
+	ReapStartRequest()
+	ReapEndRequest()
 }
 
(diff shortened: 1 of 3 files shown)


nsenter/event.go
```diff
@@ -548,6 +566,14 @@ func (e *NSenterEvent) ReceiveResponse() *domain.NSenterMessage {
 	return e.ResMsg
 }
 
+func (e *NSenterEvent) ReapStartRequest() {
```
ctalledo, line 569: I don't see this function called from anywhere; same for ReapEndRequest() below.

domain/nsenter.go
```diff
@@ -132,6 +132,8 @@ type NSenterEventIface interface {
 	SetResponseMsg(m *NSenterMessage)
 	GetResponseMsg() *NSenterMessage
 	GetProcessID() uint32
+	ReapStartRequest()
```
ctalledo, line 135: I don't see any callers for these new interface methods.

ctalledo (approved):
LG, just a couple of comments.

nsenter/event.go
rodnymolina: You're right, i ended up not needing these new methods. Will remove them and merge.

domain/nsenter.go
rodnymolina: Same as above.

PR #38, opened 2021-05-28
Fix for issue 291.

ctalledo:
These two changes fix issue 291 (i.e., launching container inside a sysbox container fails when using the latest OCI runc (v1.0.0-rc95)).

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -630,7 +630,7 @@ func (p *process) getStatus(fields []string) error {
 // pointed to by symlink. For example, if path is "/proc/self/fd/3" and
 // "/proc/self/fd/3" is a symlink to "/some/path", then this function returns
 // "/some/path". Note that "self" refers to the process struct, so we replace
-// "self" with p.pid.  The path resolution is recursive: E.g., if
+// "self" with p.pid. The path resolution is recursive: E.g., if
 // "/proc/self/fd/3" symlink points to "/proc/self/cwd", and "/proc/self/cwd"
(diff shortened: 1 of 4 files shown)


process/process.go
```diff
@@ -630,20 +630,15 @@ func (p *process) getStatus(fields []string) error {
...
 func (p *process) ResolveProcSelf(path string) (string, error) {
 
 	// TODO: this function is not capable of dealing with relative paths
-	// "../../proc/self/fd/X". It also assumes procfs is mounted on /proc, so
-	// it's not capable of dealing with scenarios where something else is mounted
-	// on /proc (rare), or where procfs is mounted at a different location (e.g.,
-	// /root/proc/self). We need to improve it to deal with these issues (likely
-	// by doing a filepath walk of each component of the path, and resolving
-	// "self" when we are sure it's under a procfs mount).
+	// "../../proc/self/fd/X". It also assumes procfs is mounted on /proc.
```
rodnymolina, line 641: I know that this is a pre-existing limitation but now that you're touching this method, what don't you just simply check if the received path is relative, and you convert it to absolute right here so that we eliminate this caveat? Or do it in the caller so that this method is always called with absolute paths.

rodnymolina (approved):
Have one suggestion that can be dealt with later on given the need to have this one merged asap.

process/process.go
ctalledo: Have to think about it, because it would be a relative path relative to the CWD and root of the process doing the mount request, so it get's tricky.

PR #39, opened 2021-06-09
Fix bug in nsenter namespace path composition.

ctalledo:
Prior to this change, sysbox-fs was creating the nsenter
paths with an extra "/" which can confuse the nsenter
process:

user/:/proc/57167/ns/user

This commit fixes it to:

user:/proc/57167/ns/user


rodnymolina (approved):
LG

PR #40, opened 2021-06-21
Handler's major refactoring to offer greater emulation flexibility and accommodate new CNIs

rodnymolina:
The following changes have been made as part of the effort to support Weave-Net and Calico CNIs within system containers. Even though these new 'applications' are only supported in the Sysbox-EE product, the refactoring required to habilitate them is also ported to sysbox-CE repos.

* Handlers are now stored in a prefix-tree instead of a dictionary. This offers greater flexibility by allowing sysbox-fs to emulate resources with 'uncertain' paths (e.g., /proc/sys/net/ipv4/neigh/*/lo).
* Handler's emulation logic has now been divided into two separate sections: one hosting the I/O methods (handlers), and another one dealing with the specific nodes being emulated (resources).
* The above point improves the handler's efficiency when dealing with concurrent access to hosts resources, as the locking primitives are now placed at the resource level (and not within the handler).
* Consolidated most of the I/O lower-level routines that interact with the host's FS (most of the duplicated code in this area is gone now).
* Convert procSysCommon handler into a general passthrough one for efficiency and consistency purposes.
* Homogenized the logging output generated by the handlers.

[ already reviewed internally ]


ctalledo (approved):
This is a subset of a PR already reviewed in Sysbox-EE ... so LG!

PR #41, opened 2021-06-25
Add config option to release seccomp-notify fd handles on container exit

ctalledo:
This commit adds a config option to sysbox-fs called "seccomp-fd-release". It
solves the problem of syscall interception not working when a process execs into
a container, forks children, and then dies, causing sysbox to close the syscall
interception handle (seccomp-notify fd) and thus breaking syscall interception
for the children processes.

When set to "cont-exit" (container exit), sysbox-fs will not close the
syscall interception handle when the associated process dies but instead will
keep it open until the associated container stops, solving the exec problem. By
default, this option is set to "proc-exit" (process exit), meaning that
sysbox-fs closes the syscall interception handle as soon as the associated
process dies (i.e., no change in the behavior prior to this change).

While setting this new config option to "cont-exit" fixes the problem described
above, the drawback that this causes kernel resources associated with that
handle to be kept around even after the associated process has died. For long
living containers where lot's of processes exec into the container, this could
be a problem so it's not recommended.  However, for short lived containers where
processes exec into the container, this should be fine.

Note that this config option is expected to be temporary because starting in
kernel >= 5.8, there is a kernel mechanism to help sysbox determine when *all*
processes associated with a given seccomp notify fd have died, thus solving the
problem described above. As a result, this option is expected to be deprecated
(or become useless) in the future.

### DIFF
--- a/seccomp/tracer.go
+++ b/seccomp/tracer.go
@@ -63,6 +63,7 @@ type SyscallMonitorService struct {
 	mts                    domain.MountServiceIface          // for mount-services purposes
 	allowImmutableRemounts bool                              // allow immutable mounts to be remounted
 	allowImmutableUnmounts bool                              // allow immutable mounts to be unmounted
+	closeSeccompOnContExit bool                              // close seccomp fds on container exit, not on process exit
 	tracer                 *syscallTracer                    // pointer to actual syscall-tracer instance
 }
 
@@ -76,7 +77,8 @@ func (scs *SyscallMonitorService) Setup(
 	prs domain.ProcessServiceIface,
 	mts domain.MountServiceIface,
 	allowImmutableRemounts bool,
-	allowImmutableUnmounts bool) {
+	allowImmutableUnmounts bool,
+	seccompFdReleasePolicy string) {
 
 	scs.nss = nss
 	scs.css = css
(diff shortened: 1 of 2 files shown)


seccomp/tracer.go
```diff
@@ -205,45 +235,77 @@ func (t *syscallTracer) seccompSessionAdd(s seccompSession) error {
...
-	if err := syscall.Close(int(s.fd)); err != nil {
-		logrus.Errorf("failed to close seccomp fd %v for pid %d: %v", s.fd, pid, err)
+	} else {
+		closeFds = []int32{fd}
 	}
 
-	logrus.Debugf("Removed session for seccomp-tracee: %v", s)
+	t.seccompSessionMu.Unlock()
+
+	if len(closeFds) > 0 {
```
rodnymolina, line 280: Why are you adding this 'if' statement? Looks like isn't needed in this logic.

rodnymolina (approved):
Have a minor question/comment. LG.

seccomp/tracer.go
ctalledo: I added it because it's possible for []closeFds to be empty, and I wanted the debug message in L301 to come out only when we closed at least one FD.

rodnymolina: got it

PR #43, opened 2021-06-29
Few fixes for relatively-minor issues

rodnymolina:
Go one by one for context.


ctalledo (approved):
LG ... the lock fix may end up being the problem we are after (?) ...

rodnymolina:
Unfortunately not I'm afraid as I don't see a direct correlation, but perhaps there might be an indirect way that connect these two ...

PR #44, opened 2021-06-29
Fix a bug related to the seccomp-fd-release config option.

ctalledo:
A prior commit added the "seccomp-fd-release" config option to sysbox-fs, which
allows sysbox-fs to support syscall interception of processes that "exec" into
the container, fork, and then die.

However there was a bug: the sysbox-fs seccomp tracer was removing the dying
process from the seccompSessionMap, thereby preventing rearming the seccomp
notification after one invocation. This change fixes this by causing the seccomp
tracer to keep the pid in the seccompSessionMap until the container dies.

### DIFF
--- a/seccomp/tracer.go
+++ b/seccomp/tracer.go
@@ -103,9 +103,8 @@ func (scs *SyscallMonitorService) Setup(
 
 // SeccompSession holds state associated to every seccomp tracee session.
 type seccompSession struct {
-	pid    uint32 // pid of the tracee process
-	fd     int32  // tracee's seccomp-fd to allow kernel interaction
-	cntrID string // container associated with the tracee process
+	pid uint32 // pid of the tracee process
+	fd  int32  // tracee's seccomp-fd to allow kernel interaction
 }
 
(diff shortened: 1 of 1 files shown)


seccomp/tracer.go
```diff
@@ -260,9 +256,13 @@ func (t *syscallTracer) seccompSessionDelete(pid uint32) {
 			}
 
 			// if the container is no longer there, or the pid being deleted is the
-			// container's init pid, we close all seccomp fds for that container
+			// container's init pid, we close all seccomp sessions for that container
```
rodnymolina, line 259: Just realized that you added this 'ContainerLookupById' instruction in the previous commit, which reduces a bit the coupling between the container-state logic and the handlers one in terms of performance. That concerned me a bit at first glance coz we already have one of these calls in the syscall-processing code-path -- see process() --, but luckily this other one that you added is executed in the context of the sessionsMonitor thread, so we are not duplicating efforts here.

rodnymolina (approved):
LG. See my comment below for future reference.

PR #53, opened 2021-08-20
Remove sysbox-fs mountpoint dir when sysbox-fs is stopped.

ctalledo:


rodnymolina (approved):
You're touching an iface signature, so pls update the UT mocks and check that all UTs are passing.

PR #55, opened 2021-09-07
Enable sys container processes to set 'trusted.overlay.opaque' xattr on files

ctalledo:
The 'trusted.overlay.opaque' xattr can only be set from the init user-namespace
when a process has CAP_SYS_ADMIN. It can't be set from within other
user-namespaces, meaning that it can't be set by the root process inside a
Sysbox container.

This causes problems when running Docker inside a Sysbox container, as described
in Sysbox issue #254.

This commit resolves this, by making Sysbox trap the *xattr() syscalls
originating inside the Sysbox container that operate on the
"trusted.overlay.opaque" attribute. Sysbox traps the syscall and performs the
operation on behalf of the process that originated the syscall. Since Sysbox
runs as root in the init user-namespace, it has the permissions to do this.

Note that this does not normally present a security risk, as processes inside
the sys container can only set the 'trusted.overlay.opaque' attribute on files
inside the container's chroot jail or files mounted into the container. In
particular, the container can never set the attribute on the overlayfs layers
that make-up the container's rootfs (unless of course the user mounts those into
the container itself (e.g., docker run --runtime=sysbox-runc -v /var/lib/docker:mnt ...),
at which point container isolation is broken in more serious ways anyway).

### DIFF
--- a/domain/nsenter.go
+++ b/domain/nsenter.go
@@ -64,31 +64,37 @@ var AllNSsButUser = []NStype{
 // by nsenterEvent class.
 //
 const (
-	LookupRequest         NSenterMsgType = "lookupRequest"
-	LookupResponse        NSenterMsgType = "lookupResponse"
-	OpenFileRequest       NSenterMsgType = "openFileRequest"
-	OpenFileResponse      NSenterMsgType = "openFileResponse"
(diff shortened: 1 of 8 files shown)


rodnymolina (approved):
nothing to add, looks great!

PR #60, opened 2021-11-01
Fix issue preventing get/listxattr() syscalls from being properly handled within a sys-container

rodnymolina:
Problem may be reproduced in scenarios where a considerable number of getxattr() and/or listxattr() syscalls are launched within a sys-container. Notice that Sysbox's proper operation isn't affected in any way during problem reproduction -- only the user process generating the syscalls may be impacted.

These syscalls are utilized to fetch the extended-attributes (xattrs) associated to any node within a file-system, and processes invoking them usually obtain this state in two steps: 1) A first syscall is made to obtain the size of the xattrs elements that the kernel is aware of for any given node, and 2) Based on the 'size' attribute obtained in 1), the user process allocates enough memory to hold these upcoming xattrs, and then it executes a second syscall to obtain the actual state from the kernel.

An important aspect to keep in mind is that the size of the xattrs returned in 2) may be a subset of the one returned by 1). This is a result of the kernel waiting till 2) to filter the xattrs that may be 'seen' by a given process based on its credentials/capabilities. Refer to xattr(7) for details.

As can be seen in the following strace capture, the problem is reproduced when the list of returned xattrs in 2) is nil:

```
llistxattr("/nix/store/nk6b2ckznjic5wj8ddw0wgdrn4mbz3lg-patch-registry-deps", NULL, 0) = 23
llistxattr("/nix/store/nk6b2ckznjic5wj8ddw0wgdrn4mbz3lg-patch-registry-deps", "", 23) = 0
```

In this case the file being queried has a 'trusted.x' attribute associated, which explains the 'size = 23' return in 1). However, the process generating the syscalls isn't holding the CAP_SYS_ADMIN capability, which explains the 'size = 0' returned in 2).

Prior to this patch Sysbox could potentially write state into the user process' address-space based on 1)'s results. The fix is to allow this to happen only when there's actual state to write as per 2)'s outcome.

As an optimization to avoid Sysbox needing to filter the xattrs to account for processes' capabilities, sysbox-fs now makes some adjustments to allow kernel's response to be the expected one.

### DIFF
--- a/nsenter/event.go
+++ b/nsenter/event.go
@@ -1010,6 +1010,28 @@ func (e *NSenterEvent) processGetxattrSyscallRequest() error {
 	p := e.ReqMsg.Payload.(domain.GetxattrSyscallPayload)
 	val := make([]byte, p.Size)
 
+	// Create a dummy 'process' struct to represent the 'sysbox-fs nsenter' process
+	// executing this logic.
+	this := e.service.prs.ProcessCreate(0, 0, 0)
+
+	// Adjust 'nsenter' process personality to match the end-user's original
+	// process.
+	if err := this.AdjustPersonality(
+		p.Header.Uid,
+		p.Header.Gid,
+		p.Header.Root,
+		p.Header.Cwd,
+		p.Header.Capabilities); err != nil {
+
+		// Send an error-message response.
+		e.ResMsg = &domain.NSenterMessage{
+			Type:    domain.ErrorResponse,
+			Payload: &fuse.IOerror{RcvError: err},
+		}
+
+		return nil
+	}
+
 	if p.Syscall == "lgetxattr" {
 		size, err = unix.Lgetxattr(p.Path, p.Name, val)
 	} else {
@@ -1070,6 +1092,28 @@ func (e *NSenterEvent) processListxattrSyscallRequest() error {
 
 	p := e.ReqMsg.Payload.(domain.ListxattrSyscallPayload)
 	val := make([]byte, p.Size)
+
+	// Create a dummy 'process' struct to represent the 'sysbox-fs nsenter' process
+	// executing this logic.
+	this := e.service.prs.ProcessCreate(0, 0, 0)
+
+	// Adjust 'nsenter' process personality to match the end-user's original
+	// process.
+	if err := this.AdjustPersonality(
+		p.Header.Uid,
+		p.Header.Gid,
+		p.Header.Root,
+		p.Header.Cwd,
+		p.Header.Capabilities); err != nil {
+
+		// Send an error-message response.
+		e.ResMsg = &domain.NSenterMessage{
+			Type:    domain.ErrorResponse,
+			Payload: &fuse.IOerror{RcvError: err},
+		}
+
+		return nil
+	}
 
 	if p.Syscall == "llistxattr" {
 		size, err = unix.Llistxattr(p.Path, val)
--- a/nsenter/event.go
+++ b/nsenter/event.go
@@ -847,7 +847,7 @@ func (e *NSenterEvent) processMountSyscallRequest() error {
 		// executing this logic.
 		this := e.service.prs.ProcessCreate(0, 0, 0)
 
-		// Adjust 'nsenter' process personality to match the end-user's original
+		// Adjust 'nsenter' process personality to match the container's original
 		// process.
 		if err := this.AdjustPersonality(
 			header.Uid,
@@ -1014,7 +1014,7 @@ func (e *NSenterEvent) processGetxattrSyscallRequest() error {
 	// executing this logic.
 	this := e.service.prs.ProcessCreate(0, 0, 0)
 
-	// Adjust 'nsenter' process personality to match the end-user's original
+	// Adjust 'nsenter' process personality to match the container's original
 	// process.
 	if err := this.AdjustPersonality(
 		p.Header.Uid,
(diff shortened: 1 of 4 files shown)


nsenter/event.go
```diff
@@ -1010,6 +1010,28 @@ func (e *NSenterEvent) processGetxattrSyscallRequest() error {
 	p := e.ReqMsg.Payload.(domain.GetxattrSyscallPayload)
 	val := make([]byte, p.Size)
 
+	// Create a dummy 'process' struct to represent the 'sysbox-fs nsenter' process
+	// executing this logic.
+	this := e.service.prs.ProcessCreate(0, 0, 0)
+
+	// Adjust 'nsenter' process personality to match the end-user's original
```
ctalledo, line 1017: nit: "end-user's" -> "container's"

seccomp/xattr.go
```diff
@@ -397,23 +412,25 @@ func (si *listxattrSyscallInfo) processListxattr() (*sysResponse, error) {
...
+	// syscall we are processing.
+	//
+	// Notice that the nsexec process executing this syscall in the container
+	// namespaces, is adopting the 'personality' of the original user process.
+	// Thus, there's no need to adjust the list of received xattrs as these ones
+	// should match those obtained if Sysbox were not in the picture.
+	//
+	// Also, bear in mind that the list of returned xattrs during the second
+	// listxattr(), may reflect a subset (or an empty set) of those reported in
+	// the first litstxattr() syscall, as it's only during the second one (i.e.,
```
ctalledo, line 424: I find it strange that the kernel only checks the process caps on the second call to listxattr() but not the first one (where size = 0). The whole point of the first listxattr() call is to get the exact number of bytes that the second call would return, so one would think that the kernel would look at the process caps in both.

This aside though, the solution you have is correct: we should use resp.Size as the "size" when writing back to the process memory, because there is no guarantee that resp.Size will equal si.Size (the kernel may return less data if for example the xattr for the file changed in between the first and second listxattr() syscalls).

ctalledo (approved):
LG ... thanks.

Did you get a chance to verify that the xattr tests in the Sysbox regression suite pass?

Is it possible to add a test to cover the scenario that leads to the bug fixed by this issue?

rodnymolina:
> LG ... thanks.
> 
> Did you get a chance to verify that the xattr tests in the Sysbox regression suite pass?

Yep, I've been playing with out xattr() testcases all the time to help me narrow down the issue.

> Is it possible to add a test to cover the scenario that leads to the bug fixed by this issue?

Yes, already did.

nsenter/event.go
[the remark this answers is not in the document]
rodnymolina: done

seccomp/xattr.go
rodnymolina: > I find it strange that the kernel only checks the process caps on the second call to listxattr() but not the first one (where size = 0). The whole point of the first listxattr() call is to get the exact number of bytes that the second call would return, so one would think that the kernel would look at the process caps in both.

Yep, I was surprised to see this too, but that's exactly what's happening as you can see below (upon arrival of the second syscall below): 

```
llistxattr("/nix/store/nk6b2ckznjic5wj8ddw0wgdrn4mbz3lg-patch-registry-deps", NULL, 0) = 23
llistxattr("/nix/store/nk6b2ckznjic5wj8ddw0wgdrn4mbz3lg-patch-registry-deps", "", 23) = 0
```

```
> github.com/nestybox/sysbox-fs/seccomp.(*listxattrSyscallInfo).processListxattr() ./sysbox-fs/seccomp/xattr.go:414 (hits goroutine(121):2 total:2) (PC: 0xd51aae)
Warning: listing may not match stale executable
   409:		// must first filter out any xattr which the process performing the syscall
   410:		// has no permissions to list.
   411:
   413:		if si.size > 0 && resp.Size > 0 {
=> 414:			if err := t.WriteRetVal(si.pid, si.addr, resp.Val); err != nil {
   415:				sysResp := t.createErrorResponse(si.reqId, syscall.ENOTSUP)
   416:				return sysResp, nil
   417:			}
   418:		}
   419:

(dlv) p si
*github.com/nestybox/sysbox-fs/seccomp.listxattrSyscallInfo {
	syscallCtx: github.com/nestybox/sysbox-fs/seccomp.syscallCtx {
		syscallNum: 195,
		syscallName: "llistxattr",
		reqId: 16518291531813794632,
		pid: 27919,
		uid: 0,
		gid: 0,
		cwd: "",
		root: "",
		processInfo: github.com/nestybox/sysbox-fs/domain.ProcessIface nil,
		cntr: github.com/nestybox/sysbox-fs/domain.ContainerIface(*github.com/nestybox/sysbox-fs/state.container) ...,
		tracer: *(*"github.com/nestybox/sysbox-fs/seccomp.syscallTracer")(0xc0000bb6e0),},
	pathFd: 0,
	path: "/nix/store/11xpmmwy95396nkhih3qc3814lqhqb8f-libunistring-0.9.10",
	addr: 140104651776544,
	size: 23,}
(dlv)

(dlv) p resp
github.com/nestybox/sysbox-fs/domain.ListxattrRespPayload {
	Val: []uint8 len: 23, cap: 24, [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],
	Size: 0,}
(dlv)
```

> This aside though, the solution you have is correct: we should use resp.Size as the "size" when writing back to the process memory, because there is no guarantee that resp.Size will equal si.Size (the kernel may return less data if for example the xattr for the file changed in between the first and second listxattr() syscalls).

Agree, that's another reason to rely on resp.Size other than the one I mentioned above.

PR #61, opened 2021-11-02
Fix for issue preventing read() operations from succeeding in a couple of procfs nodes

rodnymolina:
Read() only returns the expected value during its first invocation due to a bug in the 'caching' logic.

```
admin@test-1:~$ sudo cat /proc/sys/kernel/domainname
(none)

admin@test-1:~$ sudo cat /proc/sys/kernel/domainname

admin@test-1:~$ sudo cat /proc/sys/kernel/domainname

admin@test-1:~$
```

### DIFF
--- a/handler/implementations/utils.go
+++ b/handler/implementations/utils.go
@@ -99,8 +99,8 @@ func readFileString(
 			return 0, err
 		}
 
-		cntr.SetData(path, name, data)
 		data = val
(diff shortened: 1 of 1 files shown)


ctalledo (approved):
LG ...

(I saw and fixed this same bug in the sysbox-fs PR for handler refactoring too, but it's fine to fix it through this PR too).

PR #62, opened 2021-11-19
Remove unnecessary request offset check in sysbox-fs fuse handlers.

ctalledo:
The checks being removed are no longer necessary because the readCntrData()
and writeCntrData() functions can deal with request offsets correctly.


rodnymolina (approved):
LG

PR #63, opened 2021-11-23
Don't exit on FUSE mount errors.

ctalledo:
Sysbox-fs had a bug where it was exiting on FUSE mount errors.

This caused sysbox-fs to suddenly stop operating in a user's environment when
such an error was detected.

This change corrects this.

### DIFF
--- a/fuse/server.go
+++ b/fuse/server.go
@@ -129,7 +129,7 @@ func (s *fuseServer) Run() error {
 		fuse.DefaultPermissions(),
 	)
 	if err != nil {
-		logrus.Fatal(err)
+		logrus.Error(err)
 		return err
 	}
 


fuse/server.go
```diff
@@ -129,7 +129,7 @@ func (s *fuseServer) Run() error {
 		fuse.DefaultPermissions(),
 	)
 	if err != nil {
-		logrus.Fatal(err)
+		logrus.Error(err)
```
rodnymolina, line 132: This comes from the days we had a single fuse-server in sysbox-fs. We should have relaxed this error the moment we switched to multi-fuse-server ...

rodnymolina (approved):
LG

PR #70, opened 2022-02-17
Serialize processing of seccomp notif per thread.

ctalledo:
The sysbox-fs seccomp notification tracer (used for syscall interception) was
recently improved to parallelize processing of all received seccomp
notifications (i.e., all trapped syscalls).

This however created a problem in which sysbox-fs would receive duplicate
notifications from the kernel for a given trapped system call. It's not clear
when the kernel does this, but our debugging shows it clearly occurs.

This in turn caused sysbox-fs to disptch goroutines to process the same trapped
syscall in parallel, which then resulted in errors sent to the process that
invoked the syscall, causing it to fail. For example, doing several consecutive
"docker pulls" inside a sysbox container would sometimes fail with "failed to
register layer: error creating overlay mount to
/var/lib/docker/overlay2/<uuid>/merged: no such file or directory".

To solve this problem without affecting performance, this commit causes
sysbox-fs to serialize seccomp notif processing per kernel thread (i.e., per
seccomp notif pid). This ensures that sysbox-fs will never process two seccomp
notifs for the same thread in parallel, even if the kernel issues such
notifications. Note that this does not affect performance because threads can
only issue syscall at a time, so serializing seccomp notifs per thread on
sysbox-fs does not slow down things (except for the very small overhead of the
serialization code, which uses a golang map track pids and look them up
quickly).

### DIFF
--- a/seccomp/pidTracker.go
+++ b/seccomp/pidTracker.go
@@ -1,0 +1,106 @@
+//
+// Copyright 2019-2022 Nestybox, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package seccomp
+
+import (
+	"sync"
+)
+
+// The seccompNotifPidTracker helps serialize the processing of seccomp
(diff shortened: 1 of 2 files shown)


seccomp/pidTracker.go
```diff
@@ -0,0 +1,94 @@
...
+	"sync"
+)
+
+type pidData struct {
+	refcnt int
+	mu     sync.Mutex
+}
+
+type seccompNotifPidTracker struct {
+	mu       sync.Mutex
```
rodnymolina, line 29: Based on Lock() and Unlock() implementations, this mutex should be a RW one to reduce contention to the minimum.

rodnymolina (commented):
See optimization below.

seccomp/pidTracker.go
ctalledo: Good idea, fixed.

rodnymolina (approved):
LG

PR #71, opened 2022-02-20
Fix nsenter agent process credentials in getxattr and listxattr syscall handlers.

ctalledo:
The sysbox-fs syscall handler for getxattr and listxattr was not properly
adjusting the nsenter agent process credentials to match those of the process
that made the syscall (it was only adjusting the capabilities, but not the
uid/gid). As a result, it was getting EPERM errors when trying to access a path
in the container which was only accessible by the process that made the syscall.


rodnymolina (approved):
LG

PR #72, opened 2022-02-22
Fix sysbox-fs parsing of void * syscall arguments.

ctalledo:
Some syscalls intercepted by sysbox-fs use "void *" as arguments (e.g.,
setxattr()). Prior to this change, sysbox-fs was assuming these were strings
(i.e., nil terminated) which was causing it to misinterpret the arguments
whenever it was not bounded by a nil character in memory.

For example, while processing sexattr(), sysbox would incorrectly parse
the "value" argument to the syscalls as follows:

Want: "setxattr(): path = /mnt/tdir/tfile, name = trusted.overlay.opaque, size = 1, val = y, flags = 0"
Got: "setxattr(): path = /mnt/tdir/tfile, name = trusted.overlay.opaque, size = 1, val = yIO\xde\xc1\xfd\u007f, flags = 0"

With this commit, such arguments are now properly handled.

### DIFF
--- a/seccomp/tracer.go
+++ b/seccomp/tracer.go
@@ -554,6 +554,10 @@ func (t *syscallTracer) processMount(
 	if err != nil {
 		return t.createErrorResponse(req.Id, syscall.EPERM), nil
 	}
+
+	// Note: even though "data" is defined as a "void *" in the mount(2), we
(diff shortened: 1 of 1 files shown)


rodnymolina (approved):
LG

PR #83, opened 2023-04-19
Fix emulation of /sys/devices/virtual inside Sysbox containers

ctalledo:
See description in commits.

### DIFF
--- a/handler/implementations/passThrough.go
+++ b/handler/implementations/passThrough.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 
@@ -28,19 +29,20 @@ import (
 	"github.com/nestybox/sysbox-fs/fuse"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 //
 // Pass-through handler
 //
-// Handler for all non-emulated resources. It does a simple "passthrough" of the
-// access by entering all the namespaces of the process that is doing the I/O
-// operation and performs the access on behalf of it.
-//
-// Currently, this handler serves non-emulated resources within the /proc/sys
-// subtree, but there's nothing specific to this path in this handler's
-// implementation (see that the Path attribute is set to "*"), so this one could
-// be utilized for pass-through operations in other subtrees too.
+// Handler for all non-emulated resources under /proc/sys or /sys. It does a
+// simple "passthrough" of the access by entering the namespaces of the
+// container process that is doing the I/O operation and performs the access on
+// behalf of it. It enters the namespaces by dispatching an "nsenter agent"
+// process that enters the namespaces, performs the filesystem operation, and
+// returns the result.  Note that the nsenter agent does NOT enter the mount
+// namespace of the container process, to avoid a recursion of sysbox-fs mounts
+// /proc/sys and /sys.
 //
 
 type PassThrough struct {
@@ -62,15 +64,20 @@ func (h *PassThrough) Lookup(
 	logrus.Debugf("Executing Lookup() for req-id: %#x, handler: %s, resource: %s",
 		req.ID, h.Name, n.Name())
 
+	mountSysfs, mountProcfs, cloneFlags := checkProcAndSysRemount(n)
+
 	// Create nsenterEvent to initiate interaction with container namespaces.
 	nss := h.Service.NSenterService()
 	event := nss.NewEvent(
 		req.Pid,
 		&domain.AllNSsButMount,
+		cloneFlags,
 		&domain.NSenterMessage{
 			Type: domain.LookupRequest,
 			Payload: &domain.LookupPayload{
-				Entry: n.Path(),
+				Entry:       n.Path(),
+				MountSysfs:  mountSysfs,
+				MountProcfs: mountProcfs,
 			},
 		},
 		nil,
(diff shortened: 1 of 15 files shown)


handler/implementations/passThrough.go
```diff
@@ -483,3 +523,34 @@ func (h *PassThrough) GetResourceMutex(n domain.IOnodeIface) *sync.Mutex {
...
+	// files under /proc or /sys, the agent needs to remount these as otherwise
+	// they won't pick up the container's assigned resources (e.g., net devices,
+	// etc). We do this by having the nsenter agent in a new mount namespace so
+	// as to not mess up mounts in the host or in the container.
+
+	if strings.HasPrefix(n.Path(), "/sys/") {
+		mountSysfs = true
+	}
+
+	if strings.HasPrefix(n.Path(), "/proc/") {
```
rodnymolina, line 547: Why also for `procfs`? Have you identified a use-case that requires it?

handler/implementations/sysDevicesVirtual.go
```diff
@@ -91,7 +91,7 @@ func (h *SysDevicesVirtual) Lookup(
 		return info, nil
 	}
 
-	return n.Stat()
+	return h.Service.GetPassThroughHandler().Lookup(n, req)
```
rodnymolina, line 94: I understand the need to invoke the `passthrough` handler for nodes within the `/sys/devices/virtual` hierarchy, but why do you need to do it for the `/sys/devices/virtual` node itself, which ends up requiring complex logic to avoid nsenter accesses during container initialization? IOW, why can't you always return a dummy node for this specific node as we do (e.g.) in `sysKernel.go` or `procSys.go`?

state/containerDB.go
```diff
@@ -371,6 +373,18 @@ func (css *containerStateService) ContainerDBSize() int {
 	return len(css.idTable)
 }
 
+func (css *containerStateService) ContainerIsRegistered(id string) bool {
```
rodnymolina, line 376: As you may recall, we have been cautious to ensure that per-container i/o operations have no impact on Sysbox's registration/unregistration process/throughput (i.e., these thread groups are mostly decoupled by design). But notice that you're coupling these two functional blocks in this code, as you are stopping the global `containerStateService` every time a `passthrough` lookup() arrives...

Either way, I hope that we don't need this extra complexity -- will wait for your comments on my points above...

rodnymolina (commented):
Haven't looked at all your changes in detail, but see some initial feedback above...

handler/implementations/passThrough.go
ctalledo: I've not identified a use case for procfs yet, but I figured it would be good to add in case a handler wants the nsenter agent to mount it. Note that the code I added won't mount procfs; it's up to the handlers to set the mountProcfs flag in the nsenter request and no handler currently does it.

handler/implementations/sysDevicesVirtual.go
ctalledo: We probably don't need it for the `/sys/device/virtual` node itself, but I added the call to the passthrough handler here because it's this same handler that deals with the subdirs of `/sys/devices/virtual` (except for the subdirs that have specific handlers). 

Alternatively we could create a new handler for `/sys/device/virtual/*` and have that handler call into the passthrough handler, but not sure that's worth it.

state/containerDB.go
ctalledo: If that's an issue, I could have the handler call the container object's `IsRegistrationCompleted()` method instead. 

In that case we could probably remove the `ContainerIsRegistered()` method from the containerStateService object.

handler/implementations/sysDevicesVirtual.go
rodnymolina: No need to create a new handler. My point is that we can save the complexity associated with the need to skip the `passthrough` handler for bind-mounted resources (e.g., `/sys/devices/virtual`) during container initialization. And we do that by always returning a dummy node for lookups() targeting the `/sys/devices/virtual` node, before **and** after containers are initialized.

See an equivalent example here for `/proc/sys`:

```golang
func (h *ProcSys) Lookup(
	n domain.IOnodeIface,
	req *domain.HandlerRequest) (os.FileInfo, error) {

	logrus.Debugf("Executing Lookup() for req-id: %#x, handler: %s, resource: %s",
		req.ID, h.Name, n.Name())

	relpath, err := filepath.Rel(h.Path, n.Path())
	if err != nil {
		return nil, err
	}

	var resource = relpath

	if v, ok := h.EmuResourceMap[resource]; ok {
		if resource == "." {
			resource = "kernel"
		}
		info := &domain.FileInfo{
			Fname:    resource,
			Fmode:    v.Mode,
			FmodTime: time.Now(),
		}

		if v.Kind == domain.DirEmuResource {
			info.FisDir = true
		}

		return info, nil
	}

	return h.Service.GetPassThroughHandler().Lookup(n, req)
}
```

Btw, now that you're here, please fix this line below (cosmetic issue) by replacing 'kernel' with 'sys':

```golang
		if resource == "." {
			resource = "kernel"
		}
```

I believe an implementation such as the one I'm suggesting above, plus the existing [logic](https://github.com/nestybox/sysbox-fs/blob/a2631f69c62722c67dfd3aa97a8412b5c4db6a8a/mount/service.go#L43) to prevent recurrent sysbox-fs lookups for sysbox-fs bind-mounts, would be sufficient to address the second set of issues we previously discussed (related to recurrent i/o operations or transient issues during container init), without the need to add the extra complexity you created below (i.e., `ContainerIsRegistered()` invocations).

ctalledo: Correct; I've followed your advice and updated the changeset, please take a look again.

state/containerDB.go
ctalledo: I removed this method from the containerStateService, but left it in the container object.

handler/implementations/passThrough.go
rodnymolina: This is the main concern I have with this change: performance penalty, as we are now explicitly requesting a procfs and a sysfs mount for every incoming file operation invoking the `passthrough` handler. If we end up noticing a significant hit (which I don't really anticipate), we could always do some tuning by narrowing down the file-system paths on which we do the mount operation (e.g., there's probably no need to mount procfs if the `passthrough` handler is looking at a `/proc/sys/kernel` resource).

rodnymolina (approved):
Looks good in general, see my new comments below.

Also, as discussed, we'll need to validate these changes within sysbox-fs' enhanced handlers.

handler/implementations/sysDevicesVirtual.go
```diff
@@ -168,12 +181,13 @@ func (h *SysDevicesVirtual) ReadDirAll(
...
-	usualEntries, err := n.ReadDirAll()
-	if err == nil {
-		fileEntries = append(fileEntries, usualEntries...)
+	usualEntries, err := h.Service.GetPassThroughHandler().ReadDirAll(n, req)
+	if err != nil {
+		return nil, err
 	}
 
-	// Uniquify entries to return.
+	fileEntries = append(fileEntries, usualEntries...)
```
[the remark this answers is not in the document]
ctalledo: We do uniquify (see L191).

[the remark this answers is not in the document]
rodnymolina: oh, missed that.

PR #85, opened 2023-08-12
sysDevicesVirtual handler: use passthrough handler for /sys/devices/virtual/net/*

ctalledo:
The sysDevicesVirtual handler must enter the container namespaces for files under /sys/devices/virtual/net, since the files under these dir are namespaced resources. For other resources under /sys/devices/virtual, avoid entering the container namespaces as it's not required (the resources are not namespaced) and doing so slows down Sysbox too much.

Fixes sysbox issue 719.

### DIFF
--- a/handler/implementations/sysDevicesVirtual.go
+++ b/handler/implementations/sysDevicesVirtual.go
@@ -19,6 +19,7 @@ package implementations
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
@@ -40,6 +41,7 @@ import (
 
 type SysDevicesVirtual struct {
 	domain.HandlerBase
+	passthruNodes map[string]bool
 }
 
 var SysDevicesVirtual_Handler = &SysDevicesVirtual{
@@ -59,6 +61,9 @@ var SysDevicesVirtual_Handler = &SysDevicesVirtual{
 				Enabled: true,
 			},
 		},
+	},
+	map[string]bool{
+		"net": true,
 	},
 }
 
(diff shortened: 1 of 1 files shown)


handler/implementations/sysDevicesVirtual.go
```diff
@@ -40,6 +41,7 @@ import (
 
 type SysDevicesVirtual struct {
 	domain.HandlerBase
+	passthruNodes map[string]bool
```
rodnymolina, line 44: Instead of creating a new separate map, I would place this property (`passthruNode bool`) within the existing EmuResourceMap so that we can easily re-use this approach in other handlers in the future.

handler/implementations/sysDevicesVirtual.go
```diff
@@ -104,15 +109,19 @@ func (h *SysDevicesVirtual) Lookup(
...
-	// XXX: for non emulated resources under /sys/devices/virtual, we should
-	// ideally request the passthrough handler to perform the lookup; however
-	// this slows down the lookup and causes sysbox containers with systemd
-	// inside to fail in hosts with kernel < 5.19 (i.e., systemd takes too long
-	// to boot because for some reason it's doing lots of lookups of
-	// /sys/virtual/devices/block/loopX devices, causing it to timeout).  Instead
-	// we do the lookup at host level, which is perfectly fine except since we
-	// don't know of any files under /sys/virtual/devices which Linux exposes at
-	// host level but not inside containers.
+	// For non emulated resources under /sys/devices/virtual, we should
```
rodnymolina, line 112: We need to extend this to other file ops -- at least read() -- otherwise, some apps will continue to fail after discovering the existence of these `*/net` nodes.

rodnymolina (changes requested):
LG, but pls look at my suggested changes.

handler/implementations/sysDevicesVirtual.go
ctalledo: I am not sure: in this case we are not emulating the `/sys/virtual/devices/net` resource, so it doesn't belong in the `EmuResourceMap`. Thus it would be confusing to add it there.

handler/implementations/sysDevicesVirtual.go
ctalledo: Yep, good point, I missed that. I've added that now, thanks!

handler/implementations/sysDevicesVirtual.go
rodnymolina: Got your point. LGTM.

rodnymolina (approved):
LG

PR #86, opened 2023-11-18
Sysbox EE to CE sync

ctalledo:
Brings in several fixes and other changes from Sysbox Enterprise (used by Docker only) to Sysbox Community.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "bazil"]
```
rodnymolina, line 1: Delete this commit and the previous one (`temporarily remove bazil`) since we need relative-paths to avoid breaking recursive-github-cloning for https/ssh users.

rodnymolina (changes requested):
There are two commits that should be removed. Please check my comment above.

.gitmodules
ctalledo: Good catch, done.

PR #88, opened 2024-03-17
Fixes for /proc emulation for kernel 6.5+.

ctalledo:
Starting with Linux kernel 6.5, sysbox-fs emulation of /proc inside a container is broken, such that when the container reads files under /proc/sys, the files are empty. The reason for the breakage is unclear, but it's either a regression in the kernel or more likely an update of the FUSE protocol at kernel level which the bazil/fuse library used by Sysbox has not been updated to take into account.

The breakage occurs when Sysbox emulates the size of the files under the container's /proc to have size = 0 (as they are in the real /proc). This causes kernel 6.5+ to not read the files for some reason. The work-around is for Sysbox to report a non-zero value for the size of the files under the container's /proc. This PR does this.

### DIFF
--- a/fuse/dir.go
+++ b/fuse/dir.go
@@ -239,12 +239,16 @@ func (d *Dir) Create(
 
 	// Handler execution. 'Open' handler will create new element if requesting
 	// process has the proper credentials / capabilities.
-	err := handler.Open(ionode, handlerReq)
+	nonSeekable, err := handler.Open(ionode, handlerReq)
 	if err != nil && err != io.EOF {
 		logrus.Debugf("Open() error: %v", err)
 		return nil, nil, err
 	}
+
 	resp.Flags |= fuse.OpenDirectIO
+	if nonSeekable {
+		resp.Flags |= fuse.OpenNonSeekable
+	}
 
 	// To satisfy Bazil FUSE lib we are expected to return a lookup-response
 	// and an open-response, let's start with the lookup() one.
--- a/domain/handler.go
+++ b/domain/handler.go
@@ -95,7 +95,7 @@ type HandlerRequest struct {
 // HandlerIface is the interface that each handler must implement
 type HandlerIface interface {
 	// FS operations.
-	Open(node IOnodeIface, req *HandlerRequest) error
+	Open(node IOnodeIface, req *HandlerRequest) (bool, error)
 	Lookup(n IOnodeIface, req *HandlerRequest) (os.FileInfo, error)
 	Read(node IOnodeIface, req *HandlerRequest) (int, error)
 	Write(node IOnodeIface, req *HandlerRequest) (int, error)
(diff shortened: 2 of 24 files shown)


rodnymolina (approved):
LGTM!

fuse/dir.go
```diff
@@ -168,6 +168,13 @@ func (d *Dir) Lookup(
 	d.server.nodeDB[path] = &newNode
 	d.server.Unlock()
 
+	// Create the response
+	//
+	// XXX: how about the other fields of resp?
+	// resp.NodeID (must be unique)
+	// resp.Generation
+	resp.Attr = fuseAttrs
```
rodnymolina, line 176: I don't remember all the details but I believe there's no need to explicitly return these `fuseAttrs` as part of the `lookup()` file-operation. A `getAttr()` comes right after a `lookup()` to fetch these `fuseAttrs` that we are saving in the `NewFile` / `NewDir` instructions above.

Take a look at this code-path:

1) https://github.com/nestybox/fuse/blob/45cd9a3d884448418546d8eaa54ee7d772e576d6/fs/serve.go#L1057

2) https://github.com/nestybox/fuse/blob/45cd9a3d884448418546d8eaa54ee7d772e576d6/fs/serve.go#L1404

3) https://github.com/nestybox/fuse/blob/45cd9a3d884448418546d8eaa54ee7d772e576d6/fs/serve.go#L260

rodnymolina: In short, `file.Attrs()` instruction is indirectly utilized by several file-ops: `lookup()`, `mkdir()`, `create()`, `getAttr()`, etc.

fuse/dir.go
```diff
@@ -246,12 +246,16 @@ func (d *Dir) Create(
...
 	// process has the proper credentials / capabilities.
-	err := handler.Open(ionode, handlerReq)
+	nonSeekable, err := handler.Open(ionode, handlerReq)
 	if err != nil && err != io.EOF {
 		logrus.Debugf("Open() error: %v", err)
 		return nil, nil, err
 	}
+
 	resp.Flags |= fuse.OpenDirectIO
+	if nonSeekable {
```
rodnymolina, line 256: My understanding is that none of the procfs nodes are seekable, so what about setting these flag for all?

fuse/dir.go
rodnymolina: Technically, you can leave this new code here, but if you do, please add comments to explain what I said above (i.e., there's no real need for it taking into account bazil-fuse-lib's code structure).

ctalledo: Ok got it, but I disagree a little bit :)

I think we should return the attributes in Lookup() because it's part of the [FUSE LookupResponse](https://pkg.go.dev/bazil.org/fuse@v0.0.0-20230120002735-62a210ff1fd5#LookupResponse). And I disagree we need to add a comment, because it's part of the interface to the FUSE library. The fact that FUSE happens to call GetAttr() right after it's an internal detail of the library which could change at any time.

fuse/dir.go
ctalledo: Turns out some are; without this (i.e., if we mark all as non-seekable) the sysbox containers with systemd do not start because the systemd sysclt service fails to configure some sysctls (not sure which ones though). With this, things work as expected.

rodnymolina: interesting, didn't know...

PR #90, opened 2024-03-20
Ensure sysbox-fs nsenter processes don't leak host mount info inside containers.

ctalledo:
Background:

When a container performs an access to /proc or /sys resources emulated by
sysbox-fs, or when it runs a system call trapped by Sysbox, Sysbox sometimes
dispatches an ephemeral "nsenter" process that enters the container's
namespaces, performs an action within them, and exits.

Problem:

An issue was discovered and reported by a team at Klarna.com, whereby the
nsenter process was exposing host mount info inside the container. Since it's
common for the nsenter process to enter the container's pid namespace, it's PID
and therefore it's mounts, were visible by other processes within the
container. Thus, host mount info was "leaked" inside the container, which is not
ideal (even though it's unlikely the container could do much with it since it's
guarded by it's chroot jail and linux namespaces, particularly the user
namespace).

The underlying problem was that the nsenter process was in some cases not
entering the container's mount namespace, because it wanted to perform
mounts of procfs and sysfs without the container noticing. However, by not
entering the container's mount namespaces, it was carrying with it the
host mounts it inherits from its parent process (sysbox-fs).

Solution:

The solution is for the nsenter process to always enter the container's mount
namespace; this way it let's go of it's parent process mounts and only sees the
container's mounts (i.e., no more host mount leakage in the container). If the
nsenter process needs to do some mounts of it's own, then after having entered
the container's mount namespaces, it will unshare its mount namespace so that it
can perform these mounts without affecting the container. Note that even though
other container processes can see the PID of the ephemeral nsenter process, they
won't be able to access any new mounts done by this process and won't have
permission to enter its mount namespace either.

Special thanks to Gabriele Diener at Klarna.com for reporting this issue.

### DIFF
--- a/nsenter/utils.go
+++ b/nsenter/utils.go
@@ -17,21 +17,103 @@
 package nsenter
 
 import (
+	"os"
+	"strings"
+
 	"golang.org/x/sys/unix"
 )
 
-func processPayloadMounts(mountSysfs, mountProcfs bool) error {
-	var flags uintptr = unix.MS_NOSUID | unix.MS_NOEXEC | unix.MS_NODEV | unix.MS_RELATIME
+type payloadMountsInfo struct {
+	sysfsMountpoint  string
+	procfsMountpoint string
+	cleanup          func()
+}
+
+// This function runs when the sysbox-fs nsenter helper process requests to
+// mount procfs or sysfs.
+//
+// Note that the container process should do so from an unshared mount
+// namespace, such that the mounts are NOT visible inside the container
+// (otherwise container processes will see the nsenter process mounts which are
+// inherited from sysbox-fs mounts, thus leaking host info into the container).
+//
+// Note also that the nsenter process mounts the real procfs and sysfs, not the
+// sysbox-fs emulated ones. That's because the nsenter process is not under
+// seccomp-notify intercepts on mount syscalls as the container processes are.
+func processPayloadMounts(mountSysfs, mountProcfs bool) (*payloadMountsInfo, error) {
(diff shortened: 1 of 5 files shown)


nsenter/utils.go
```diff
@@ -17,21 +17,103 @@
...
+		sysfsMountpoint, err = os.MkdirTemp("/", ".sysbox-sysfs-")
+		if err != nil {
+			return nil, err
+		}
+		if err = unix.Mount("sysfs", sysfsMountpoint, "sysfs", flags, ""); err != nil {
+			os.RemoveAll(sysfsMountpoint)
+			return nil, err
+		}
+	}
+	cleanupSysfs := func() {
```
rodnymolina, line 76: Shouldn't these cleanup routines be within the `if mountSyfs` and `if mountProcfs` code blocks? Why do you want to return a cleanup() function when no `mount()` has been made?

rodnymolina (commented):
Great solution! Most of it makes perfect sense but I'm struggling a bit with `processPayloadMounts()` function. WIP.

rodnymolina (approved):
Never mind, I got it now.

nsenter/utils.go
ctalledo: Yes exactly, the cleanups only work when sysfs or procfs are mounted.

The reason I needed them is because the cleanup (e.g., removal of the sysfs mountpoint) needs to occur after the event handling (done by the caller of procssPayloadMounts) has taken place.

PR #91, opened 2024-04-07
Add support for x86 syscalls in seccomp tracer

joanbm:
Sysbox recently allowed x86 syscalls, thus allowing x86 (32-bit) applications to run inside Sysbox.
(https://github.com/nestybox/sysbox-runc/commit/085502643ea5281652c6984eed9797872f22698a).

However, since the same syscall can have different numbers on x86 and x86_64 (e.g. `mount` is NR=165 on x86_64 but NR=21 on x86), the seccomp tracer needs to look at the architecture to decode the syscall. But currently, it decodes all syscalls as if they were x86_64 syscalls, so running the following command on an amd64 host fails:

    # docker run --rm -it --runtime=sysbox-runc i386/alpine \
          mount -t tmpfs none /mnt
    mount: mounting none on /mnt failed: Invalid argument

With a warning being printed in the journal:

    Unsupported syscall notification received (21) on fd 23, [...]

Running more complex applications such as `dockerd` will also fail.

This commit modifies the tracer to keep separate tables for x86 and x86_64 and to decode incoming syscalls according to their architecture.

### DIFF
--- a/seccomp/tracer.go
+++ b/seccomp/tracer.go
@@ -112,6 +112,11 @@ func (scs *SyscallMonitorService) Setup(
 	}
 }
 
+type seccompArchSyscallPair struct {
+	archId    libseccomp.ScmpArch
+	syscallId libseccomp.ScmpSyscall
+}
+
 // SeccompSession holds state associated to every seccomp tracee session.
 type seccompSession struct {
 	pid    uint32 // pid of the tracee process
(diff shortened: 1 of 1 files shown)


seccomp/tracer.go
```diff
@@ -148,19 +168,27 @@ func newSyscallTracer(sms *SyscallMonitorService) *syscallTracer {
...
-			logrus.Warnf("Seccomp-tracer initialization error: unknown syscall (%v).",
-				syscall)
-			return nil
+	nativeArchId, err := libseccomp.GetNativeArch()
+	if err != nil {
+		logrus.Warnf("Seccomp-tracer initialization error: Error obtaining native architecture")
+		return nil
+	}
+
+	for archId, sycalls := range getSupportedCompatibleSyscalls(nativeArchId) {
```
ctalledo, line 177: nit: sycall -> syscalls

ctalledo (approved):
Thanks @joanbm for catching this problem and fixing it, much appreciated.

Looks good, I only have a tiny suggestion (see comments).

joanbm:
Thanks for the review and catching the typo, PR updated!

PR #92, opened 2024-05-07
Fix to allow access to /proc/sys/kernel/shm* files inside Sysbox containers.

ctalledo:
Even though the /proc/sys/kernel/shm* files (to control shared memory limits) are
namespaced in the kernel via the IPC namespace, the kernel only allows true root
to write to these files. Therefore, a sysbox container root user is not allowed
to write to these files. For example:

```
$ docker run --runtime=sysbox-runc --rm -it alpine sh
#/ echo 1024 > /proc/sys/kernel/shmmax
error: permission denied
```

To overcome this, improve the sysbox-fs procSysKernel handler to service
accesses to these files without entering the Sysbox container's user-namespace.
This also required corresponding changes to the passthrough handler.

### DIFF
--- a/handler/implementations/procSysKernel.go
+++ b/handler/implementations/procSysKernel.go
@@ -381,6 +381,13 @@ func (h *ProcSysKernel) Open(
 
 	case "printk":
 		return false, nil
+
+	case "shmall":
+		fallthrough
+	case "shmmax":
+		fallthrough
+	case "shmmni":
+		return h.Service.GetPassThroughHandler().OpenWithNS(n, req, domain.AllNSsButUser)
 	}
 
 	// Refer to generic handler if no node match is found above.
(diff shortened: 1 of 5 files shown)


handler/implementations/procSysKernel.go
```diff
@@ -492,6 +506,18 @@ func (h *ProcSysKernel) Write(
...
+		fallthrough
+	case "shmmax":
+		fallthrough
+	case "shmmni":
+		// The kernel only allows true root to write to /proc/sys/kernel/shm*.
+		// Root in the container's user-namespaces is not allowed to modify these
+		// values, even though they are namespaced via the IPC namespace.
+		// Therefore ask the passhthrough handler to enter all namespaces except
+		// the user-ns, as otherwise we get permission denied.
+		return h.Service.GetPassThroughHandler().WriteWithNS(n, req, domain.AllNSsButUser)
```
rodnymolina, line 520: Do we really need to push this value to kernel? Can't we emulate it instead as we do with other values avoiding writing the value to the kernel? I'm thinking about potential memory DoS attacks here...

rodnymolina (commented):
Got a question.

handler/implementations/procSysKernel.go
ctalledo: Yes, we want to push this value to the kernel because (1) it's namespaced via the IPC ns (so will only affect the container), and (2) it allows a user to restrict the amount of shared memory that a container can use. So in fact this helps mitigate DoS attacks.

rodnymolina (approved):
LGTM

PR #97, opened 2024-10-09
Fix for panic when reading from process addr space.

ctalledo:
Fixes a "panic: runtime error: index out of range [0] with length 0" in the sysbox-fs code that reads from a container process' memory (see sysbox issue

In function readProcessMem() in seccomp/memParsesIOvec.go, the code was assuming a data buffer slice would have at least one element, but in cases where the read size is 0, the slice had no elements, thus causing the panic.

This change ensures that we only call readProcessMem() when the data buffer has size > 0.

As a bonus, this commit also refactors / simplifies the readProcessMem() function by replacing the two-dimensional array "local" to a one-dimensional array (only one dimension is actually needed).


rodnymolina (approved):
LGTM

PR #99, opened 2024-10-18
Emulate /proc/sys/kernel/random/uuid inside Sysbox containers.

ctalledo:
For some unknown reason, accessing /proc/sys/kernel/random/uuid from inside a Sysbox container was always returning the same value (instead of returning a different value on each read). See sysbox issue #785.

I don't know why this is, but it must be due to some kernel bug triggered when the read of the uuid comes from the sysbox-fs passthrough handler, which enters the container namespaces and then performs the read.

Fix this by emulating /proc/sys/kernel/random/uuid via a new handler. The handler simply reads the same file from the host kernel, which as expected returns a different value every time.


rodnymolina (approved):
LGTM

PR #101, opened 2024-12-05
Resolve /proc/thread-self/* in syscall interception code.

ctalledo:
When intercepting system calls that use filepaths as arguments, those paths may be regular paths or may be paths under /proc (e.g., "/proc/self/fd/<num>", "/proc/self/task/<tid>/fd/<num>", or "/proc/thread-self/fd/<num>"). Sysbox was dealing properly with the first two but not with the last one (i.e., "/proc/thread-self/*").

This went unnoticed until a recent change in the OCI runc (commit 8e8b136c4923ac33567c4cb775c6c8a17749fd02) where it's now using "/proc/thread-self/<path>" in mount system calls it issues. When that new OCI runc ran inside a Sysbox container, it failed with an error such as:

"error mounting "proc" to rootfs at "/proc": mount src=proc, dst=/proc, dstFd=/proc/thread-self/fd/8, flags=0xe: no such file or directory: unknown."

See https://github.com/nestybox/sysbox/issues/879 for further info.

This commit fixes this by updating Sysbox to deal properly with syscalls that have "/proc/thread-self/*" in their arguments.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -639,9 +639,13 @@ func replaceProcSelfWithProcPid(path string, pid uint32, tid uint32) string {
 
 	pidMatch := regexp.MustCompile(`^/proc/self/(.*)`)
 	tidMatch := regexp.MustCompile(`^/proc/self/task/[0-9]+/(.*)`)
+	tidMatch2 := regexp.MustCompile(`^/proc/thread-self/(.*)`)
 
 	repl = fmt.Sprintf("/proc/self/task/%d/${1}", tid)
 	p = tidMatch.ReplaceAllString(path, repl)
+
+	repl = fmt.Sprintf("/proc/self/task/%d/${1}", tid)
+	p = tidMatch2.ReplaceAllString(p, repl)
 
 	repl = fmt.Sprintf("/proc/%d/${1}", pid)
 	p = pidMatch.ReplaceAllString(p, repl)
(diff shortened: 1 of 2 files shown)


process/process.go
```diff
@@ -639,10 +639,14 @@ func replaceProcSelfWithProcPid(path string, pid uint32, tid uint32) string {
 
 	pidMatch := regexp.MustCompile(`^/proc/self/(.*)`)
 	tidMatch := regexp.MustCompile(`^/proc/self/task/[0-9]+/(.*)`)
+	tidMatch2 := regexp.MustCompile(`^/proc/thread-self/(.*)`)
 
 	repl = fmt.Sprintf("/proc/self/task/%d/${1}", tid)
 	p = tidMatch.ReplaceAllString(path, repl)
 
+	repl = fmt.Sprintf("/proc/self/task/%d/${1}", tid)
```
rodnymolina, line 647: If I understand this properly, does it make sense to run lines 645 and 648 in sequence? Shouldn't we run either 645 or 648?

ctalledo: It's going to be one or the other anyway because the regexp don't overlap. In other words, either the `tidMatch` regex hits or the `tidMatch2` regex hits, but both can never hit (i.e., the first regex matches paths starting with `/proc/self/task`, the second with `/proc/thread-self/`). Thus I kept it that way, instead of adding an extra `if-else` clause.

rodnymolina: Got it.

rodnymolina (approved):
LGTM

PR #103, opened 2025-01-16
Fix for issue 854

ctalledo:
Delay PathAccess check in umount syscall processing.
    
In the `umount` syscall processing, rather than performing the `PathAccess()` check upfront (to check if the process has the required permissions to access the directory being unmounted), delay it until we know it refers to a sysbox-fs managed mount or an immutable mount.
    
This means that for other mountpoints (i.e., those not managed by sysbox-fs), we let the kernel do the path access check since it's the kernel that processes those mounts anyway.
    
Why this change? because this allow us to work-around a problem where while processing the `umount` syscall, `PathAccess()` unexpectedly fails on a FUSE-backed mount inside the container (see [Sysbox issue #854](https://github.com/nestybox/sysbox/issues/854)). The failure is caused by the FUSE backed mount not showing up properly under `/proc/<pid>/root/<path-to-fuse-mountpoint>` (where `<pid>` is the container's process ID in the Sysbox pid namespace).
    
Turns out that for some reason, for FUSE-backed mountpoints, the kernel only shows the mountpoint under `/proc/<pid>/root/<path-to-fuse-mountpoint>` when it's read from within the user and mount namespaces of the container. Thus it fails when accessed by sysbox-fs directly. This behavior does not reproduce on other mountpoint types, which is why we didn't detect this problem before (e.g., I tried with a tmpfs mount inside the container, and verified that `/proc/<pid>/root/<path-to-tmpfs-mountpoint>` works fine).

So in short, this PR does an "easy" work-around for a problem whose "real" solution would require sysbox-fs to enter the namespaces of the container and perform the PathAccess() check from within.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -712,9 +712,16 @@ func (p *process) ResolveProcSelf(path string) (string, error) {
 		}
 
 		// path is a symlink, resolve it
-		currPath, err = os.Readlink(currPath)
+		link, err := os.Readlink(currPath)
 		if err != nil {
 			return "", err
+		}
+
+		if filepath.IsAbs(link) {
+			currPath = link
+		} else {
+			// the link is relative, resolve it
+			currPath = filepath.Clean(filepath.Join(filepath.Dir(currPath), link))
 		}
 	}
 
--- a/seccomp/tracer.go
+++ b/seccomp/tracer.go
@@ -714,7 +714,20 @@ func (t *syscallTracer) processUmount(
 	// update it in case it requires path resolution.
 	umount.Target, err = process.PathAccess(umount.Target, 0, true)
 	if err != nil {
-		return t.createErrorResponse(req.ID, err), nil
+		if err == syscall.ENOENT {
+			// XXX: in some cases PathAccess() will unexpectedly hit an ENOENT when
+			// walking the path. We've seen this occur on user-created FUSE mounts
+			// inside the container, where the path walk to the FUSE mountpoint
+			// does not work unless we enter the container's user and mount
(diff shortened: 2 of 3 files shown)


rodnymolina (commented):
Thanks for the great explanation of the issue. Conceptually, this workaround would fix the problem and improve our umount() I/O throughput.

Unfortunately, I see a potential issue derived from the fact that the PathAccess() method not only verifies permissions but also adjusts/resolves the target paths.

For example, if a user were to create an absolute softlink to the container's `/proc/sys` path and attempts to umount it, then, since we are deferring the softlink resolution till we identify the target as a sysbox-fs managed resource, we would skip the protections that we have in place if the user has the `allowImmutableUnmounts` flag set to "true" (which is the default value). Consequently, we would hand out the umount() instruction to the kernel, allowing the operation to succeed.

ctalledo:
> Unfortunately, I see a potential issue derived from the fact that the PathAccess() method not only verifies permissions but also adjusts/resolves the target paths.
> 
> For example, if a user were to create an absolute softlink to the container's `/proc/sys` path and attempts to umount it, then, since we are deferring the softlink resolution till we identify the target as a sysbox-fs managed resource, we would skip the protections that we have in place if the user has the `allowImmutableUnmounts` flag set to "true" (which is the default value). Consequently, we would hand out the umount() instruction to the kernel, allowing the operation to succeed.

Thanks @rodnymolina for the response; yes that's a valid point and therefore a mistake in the PR. I've re-worked it now to avoid the scenario you described, please take a look (it's actually a simpler change now).

rodnymolina (approved):
LGTM

PR #106, opened 2025-07-15
Changes in sysbox-fs to use latest Bazil (fuse) lib.

ctalledo:
Sysbox-fs uses a forked and slightly modified version of the golang Bazil lib to implement file-system in user-space (FUSE) functionality.

The fork was recently updated, so have sysbox-fs point to the latest version. This also requires some minor changes in sysbox-fs to adjust to changes in the Bazil lib.

Tested the full Sysbox test suite on this change to make sure everything works as expected.


ctalledo:
Merging as Rodny is on vacation.

PR #113, opened 2025-12-30
fix: FuseServerService.CreateFuseServer don't block forever when Server.Run failed

okhowang:
mitigate https://github.com/nestybox/sysbox/issues/981


ctalledo (approved):
LGTM, thanks @okhowang

PR #115, opened 2026-01-19
Port latest changes from Sysbox EE -> CE

ctalledo:
Main change is trapping the `openat2` system-call, in order to enable latest versions of `runc` (used by Docker v28+) to work inside Sysbox containers.


ctalledo:
Rodny has already reviewed these in Sysbox-EE, so merging without code-review here.

PR #118, opened 2026-04-06
sysbox-fs/seccomp: add CIFS/SMB3 mount proxy

### LINKED ISSUE #856: Permission Denied: mount of cifs inside sysbox container not working
I have permission issues with mounting SMB share using `mount -t cifs` and without sysbox its working well with right capabilities, but with sysbox it throws permission error, probably sysbox while intercepting the mount calls denying the mount of cifs? 

With Sysbox:
with-sysbox

Without Sysbox its working well.

I did some research and read other threads and found @ctalledo recommending this

> In the meantime, a work-around in order to mount a cifs-backed volume into a Sysbox container would be to configure Docker in [userns-remap mode](https://docs.docker.com/engine/security/userns-remap/). This way Sysbox won't need to use shiftfs anymore.
> 
> If you want to do this, add the "userns-remap" line to the `/etc/docker/daemon.json` file

but I cant use this since my infra relies on network_mode: host, and with userns-remap, host network is not possible since it affects even the runc containers. 

Is there a workaround for this today?

My requirement comes from having a file system that needs mandatory file locking, and I ended up using SAMBA.

VitorRamos:
Extend processNfsMount to also handle "cifs" and "smb3" filesystem types. The existing NFS proxy already does exactly what CIFS needs: re-issue mount(2) via nsenter as host root (AllNSsButUser).

- mount.go: add "cifs", "smb3" to the processNfsMount dispatch case; update function comment; use m.FsType in the debug log

Tested: ubuntu:24.04, cifs-utils 7.0, guest share, vers=3.1.1

Fixes [#856](https://github.com/nestybox/sysbox/issues/856)

### DIFF
--- a/seccomp/mount.go
+++ b/seccomp/mount.go
@@ -75,7 +75,7 @@ func (m *mountSyscallInfo) process() (*sysResponse, error) {
 			return m.processSysMount(mip)
 		case "overlay":
 			return m.processOverlayMount(mip)
-		case "nfs":
+		case "nfs", "cifs", "smb3":
 			return m.processNfsMount(mip)
 		}
 	}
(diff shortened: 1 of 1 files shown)


VitorRamos:
Hi,

We’ve been running into some SMB/CIFS mount limitations when using Sysbox. We’re currently testing it in our Coder setup (https://coder.com/docs/admin/templates/extending-templates/docker-in-workspaces), and so far this has been the main blocker.

Would this kind of change be something you’d consider accepting?

Thanks!

ctalledo (approved):
LGTM, thanks @VitorRamos !

